PR #310, opened 2015-01-02
gui: ignore (a|A)ccesib(ility|le) words

jfilak:
I am reluctant to use single "accesib" because I am afraid of false
negative results.

Related: rhbz#1175720


mmilata:
Merged, thank you!

PR #323, opened 2015-01-28
fix several memory leaks

jfilak:
I discovered these leaks while chasing a bug related to rhbz#1163579.

The most interesting one is the missing semicolon in src/lib/file_obj.c


jfilak:
I've merged this pull request from command line.

PR #326, opened 2015-02-04
reporter-mantisbt several fixes

mhabrnal:


mhabrnal:
Thank you for the good point. i've addressed your comments.

PR #334, opened 2015-03-03
ignored words: ignore lxqt-openssh-askpass

lkundrak:
This was seen in the environment:
SSH_ASKPASS=/usr/libexec/openssh/lxqt-openssh-askpass


jfilak:
Thanks!

PR #343, opened 2015-04-15
RHEL7: copy file chown

jfilak:
Needed to avoid race-conditions between copy & chown.

Related: #1211835


src/lib/concat_path_file.c
```diff
@@ -57,3 +57,28 @@ char *concat_path_basename(const char *path, const char *filename)
...
+
+    if (NOT_PRINTABLE(*str) || *str =='/' || (*str == '\0' && *(str-1) == '.'))
+        return false;
+    ++str;
+
+    if (NOT_PRINTABLE(*str) || *str =='/' || (*str == '\0' && *(str-1) == '.' && *(str-2) == '.'))
+        return false;
+    ++str;
+
+    for (unsigned i = 0; *str != '\0' && i < 61; ++str, ++i)
```
sorki: This for should be enough ^^, you don't have to check for /, ./ , ../ cases separately. 

Also why there's a limit of 61 characters?

src/lib/dump_dir.c
```diff
@@ -111,15 +112,52 @@ static bool exist_file_dir(const char *path)
     return false;
 }
 
+
+int secure_openat_read(int dir_fd, const char *pathname)
```
sorki: This imo deservers a docstring.

sorki:
Went through the code and it seems to mitigate time of check / time of use race conditions and improve malicious path handling. Ack.

src/lib/concat_path_file.c
jfilak: This code need more comments, all the security patches I've developed in the last two weeks need documentation.

Ad 61:
This function accepts only strings 64 characters long. 61 + 3 above. 

Ad the for loop:
0. you are right but I'm not going to rewrite it because:
1. this function needs to be fast
2. check for '.' make sense only for the first two characters
3. searching for '.' in the loop would be redundant because

PR #346, opened 2015-04-16
Symlink vulnerabilities and race conditions

jfilak:


sorki:
Typo in commit message of 5797c06 - "rebost".

sorki:
Otherwise fine.

Should probably also include hardening patches from rhel7_copy_file_chown branch and vice-versa if applicable (dump proc data patch).

mbrysa:
Thank you!

PR #347, opened 2015-04-22
report client: provide cpio log when unpacking fails

mhabrnal:
Related to rhbz#1169774 rhbz#1213485 rhbz#1036918


sorki:
Please wrap lines to 80 chars and squash the "fix -" commit into previous one. Otherwise fine.

mhabrnal:
rebased, I spread it into two commits because the "fix -" commit is not related with the bugzillas, It is another bug we have found.

PR #349, opened 2015-05-12
report client: check owner of /var/cache/abrt-di when unpacking fails

mhabrnal:
If unpacking of debuginfo fails and '/vat/cache/abrt-di' is not owned by
abrt.abrt the client provides a solution how to fix the issue.

Related to rhbz#1213485

### DIFF
--- a/src/client-python/reportclient/debuginfo.py
+++ b/src/client-python/reportclient/debuginfo.py
@@ -23,6 +23,7 @@
 
 import sys
 import os
+import pwd
 import time
 import errno
 import shutil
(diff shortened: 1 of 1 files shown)


src/client-python/reportclient/debuginfo.py
```diff
@@ -381,6 +383,15 @@ def download(self, files, download_exact_files=False):
                 if unpack_result == RETURN_FAILURE:
                     # recursively delete the temp dir on failure
                     print(_("Unpacking failed, aborting download..."))
+
+                    s = os.stat(self.cachedir)
+                    user = pwd.getpwuid(s.st_uid)[0]
+                    group = grp.getgrgid(s.st_gid)[0]
```
mbrysa: These two raise a `KeyError` if the uid/gid is not found.

mbrysa:
Thank you!

PR #355, opened 2015-05-27
RHEL7: ureport: add functionality to use consumer certificate

mhabrnal:
reporter-ureport uses consumer certificate instead of entitlement certificate,
if the rhsm authentication is enabled. Also uses
'https://cert-api.access.redhat.com/rs/telemetry/abrt' to report those
autenticated uReports by default.
    
Related to rhbz#1223805
    
### DIFF
--- a/src/lib/ureport.c
+++ b/src/lib/ureport.c
@@ -31,13 +31,11 @@
 
 #define BTHASH_URL_SFX "reports/bthash/"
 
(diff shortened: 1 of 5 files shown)


mhabrnal:
I've addressed @jfilak comments and corrected a comment  in ureport.conf.

jfilak:
I pushed these two commits without any changes:
34f32d6 testsuite: fix test for ureport
d8459ff ureport: add functionality to use consumer certificate

PR #359, opened 2015-05-28
RHEL7: RHTSupport: include reported_to in Support cases

mhabrnal:
We want to provide GSS with an URL to ABRT server.

Related to rhbz#1197108


jfilak:
8ceb10a RHTSupport: include reported_to in Support cases

PR #369, opened 2015-06-23
support new dump dir element 'tid'

jfilak:
Do not include it in bug reports.


mhabrnal:
Looks good for me, thanks!

PR #378, opened 2015-09-16
Use Red Hat Certificate Authority to make rhsm cert trusted

mhabrnal:


mhabrnal:
I've addressed your comment and rebased the pull request.

PR #380, opened 2015-09-30
run_event: rewrite event rule parser

sorki:
Because '# there has to be a comment here, otherwise'
sounds like 'I am too lazy to fix my parser'

This fixes ^^ and allows newlines in command blocks.


jfilak:
Thank you! I've added a new commit that fixes problems I found during review. This commit should be amended to your commit once we both agree on changes I made.

The tests look good but one allowed EVENT definition form is missing:
```
"EVENT=report_Mailx      reporter-mailx"
```
(from /etc/libreport/events.d/mailx_event.conf)

sorki:
Thanks for the review. I've added tests/rules/oneline to account for missing form.

jfilak:
I am sorry, I haven't tested my changes and we have an invalid free error now. This patch should fix it:

```
diff --git a/src/lib/run_event.c b/src/lib/run_event.c
index 11c4f25..e11b70b 100644
--- a/src/lib/run_event.c
+++ b/src/lib/run_event.c
@@ -256,22 +256,22 @@ GList *load_rule_list(GList *rule_list,
             log_parser("found include");
 
             /* "include GLOB_PATTERN" */
-            line = skip_whitespace(line + strlen("include"));
+            const char *p = skip_whitespace(line + strlen("include"));
 
             const char *last_slash;
             char *name_to_glob;
-            if (*line != '/'
+            if (*p != '/'
              && (last_slash = strrchr(conf_file_name, '/')) != NULL
             )
                 /* GLOB_PATTERN is relative, and this include is in path/to/file.conf
                  * Construct path/to/GLOB_PATTERN:
                  */
-                name_to_glob = xasprintf("%.*s%s", (int)(last_slash - conf_file_name + 1), conf_file_name, line);
+                name_to_glob = xasprintf("%.*s%s", (int)(last_slash - conf_file_name + 1), conf_file_name, p);
             else
                 /* Either GLOB_PATTERN is absolute, or this include is in file.conf
                  * (no slashes in its name). Use unchanged GLOB_PATTERN:
                  */
-                name_to_glob = xstrdup(line);
+                name_to_glob = xstrdup(p);
 
             glob_t globbuf;
             memset(&globbuf, 0, sizeof(globbuf));
```

sorki:
Fixed, thanks.

jfilak:
I think you need to implement this contract from `src/lib/report_event.c`:
```All subsequent lines which start with space or tab form one rule.```

The current code interprets the following lines as a single EVENT:
```
EVENT=clumsy_event
    echo 'I am dodo'
import dummy_event_file.conf
```

**Comment by sorki** [comment]
Fixed. Added comment and tests/rules/include_right_after_event test

**Comment by jfilak** [comment]
Fantastic. Thank you!

Just last minor thing. `$ TESTSUITEFLAGS="-v 72 73 74 75" make maintainer-check-valgrind` fails and this patch fixes it:

```
diff --git a/tests/load_rule_list.at b/tests/load_rule_list.at
index 75b10e1..f436358 100644
--- a/tests/load_rule_list.at
+++ b/tests/load_rule_list.at
@@ -69,6 +69,7 @@ check(const char *input,
       const char ** expected_conditions) {
 
     GList *rule_list = load_rule_list(NULL, input, 0);
+    GList *bck_list = rule_list;
 
     int cond_id = 0;
     assert(rule_list != NULL);
@@ -90,7 +91,7 @@ check(const char *input,
     // check that there are no more conditions in expected_conditions
     assert(expected_conditions[cond_id] == NULL);
 
-    free_rule_list(rule_list);
+    free_rule_list(bck_list);
 }
 
 int main(void)
@@ -119,6 +120,7 @@ check(const char *input,
       const char *expected_command) {
 
     GList *rule_list = load_rule_list(NULL, input, 0);
+    GList *bck_list = rule_list;
 
     int cond_id = 0;
     assert(rule_list != NULL);
@@ -131,7 +133,7 @@ check(const char *input,
         rule_list = rule_list->next;
     }
 
-    free_rule_list(rule_list);
+    free_rule_list(bck_list);
 }
 
 int main(void)
```

sorki:
Fixed!

PR #384, opened 2015-10-06
bugzilla: don't attach build_ids

sorki:
Bad indent corrected


jfilak:
Sorry, I forgot to tell you to fix this in `src/plugins/mantisbt_format.conf` too.

PR #388, opened 2015-10-20
Changes needed for the integration of ABRT in QE workflows

jfilak:
We need to be able to attach the URL of the uploaded problem data to the relevant micro-report. Here are the problems addressed by this pull request.

1. reporter-upload does not save its report results in reported_to file
2. reporter-ureport does not know 'url' micro-report attachment type [abrt/faf#467]

How should be this pull request used in the integration efforts:
```
EVENT=notify
  reporter-ureport
  EC=$?

  # See  EXIT_STOP_EVENT_RUN in /usr/include/libreport/internal_libreport.h
  if [ $EC -eq 70 ]; then
   echo "Already reported problem:"
    cat reported_to
    exit 0
  fi

  if [ $EC -ne 0 ]; then
    exit $EC
  fi

  reporter-upload -u ftp://qe.example.org/abrt-reports || exit 1
  reporter-ureport -r "upload" -L "URL" -T "url" || exit 1
```


jfilak:
ABRT integration test case (not pushed yet):
https://github.com/abrt/abrt/commit/3c3544be82feed8d5fb163c69ae0872ea68a0e23

jfilak:
Do not merge this pull request. The lines added to reported_to file can contain usernames and passwords:
```
upload: URL=scp://root:redhat@localhost/tmp/tmp.RaYnVplNNv/target/problem_dir.tar.gz
```

I need to either ask curl to return URL without credentials or remove everything between `://` and `@`.

jfilak:
I've fixed the problem with credentials and added the following commit to abrt:
https://github.com/abrt/abrt/commit/4b307ccfb424d0a0a66e180544b7498ce026195e

jfilak:
This pull request can be reviewed and merged.

src/plugins/reporter-upload.c
```diff
@@ -33,32 +32,48 @@ static char *ask_url(const char *message)
...
 {
-    int result = 1; /* error */
+    post_state_t *state = new_post_state(POST_WANT_ERROR_MSG);
+    state->username = getenv("Upload_Username");
+    char *password_inp = NULL;
+    if (state->username != NULL && state->username[0] != '\0')
+    {
+        /* Load Password only if Username is configured, it doesn't make */
+        /* much sense to load Password without Username. */
+        state->password = getenv("Upload_Password");
```
sorki: What if user configures scp to use keys? There might be username set but no password is needed. Also using CURL to upload stuff via SCP is not good at all, would be much better to just use SCP so user can configure it as usual and not fiddle with CURL/abrt specific configuration.

src/plugins/reporter-ureport.c
```diff
@@ -90,12 +95,23 @@ int main(int argc, char **argv)
...
+                          _("use REPORT_RESULT_TYPE when looking for FIELD in reported_to (used only with -V)")),
+        OPT_STRING('T', "type", &attach_type, "TYPE",
+                          _("attach DATA as ureporte attachment TYPE (used only with -v|-V)")),
         OPT_END(),
     };
 
     const char *program_usage_string = _(
         "& [-v] [-c FILE] [-u URL] [-k] [-t SOURCE] [-h CREDENTIALS]\n"
-        "  [-A -a bthash -B -b bug-id -E -e email -O -o comment] [-d DIR]\n"
+        "  [-A -a bthash -B -b bug-id -E -e email -O -o comment -R -r url] [-d DIR]\n"
```
sorki: -V -r ?

src/plugins/reporter-upload.c
jfilak: Good point! I will try to rework this part later on: https://github.com/abrt/libreport/issues/389

src/plugins/reporter-ureport.c
jfilak: Eek, this is a left-over from the very first version of this commit. Good catch. Thanks!

jfilak:
I corrected the documentation issues found by @sorki

sorki:
Good to go from my side, thanks.

mhabrnal:
Looks good to me, thanks.

PR #398, opened 2015-12-15
dd: make function uid_in_group() public

mhabrnal:
Related to rhbz#1256705

### DIFF
--- a/src/include/internal_libreport.h
+++ b/src/include/internal_libreport.h
@@ -906,6 +906,15 @@ int delete_dump_dir_possibly_using_abrtd(const char *dump_dir_name);
 #define steal_directory libreport_steal_directory
 struct dump_dir *steal_directory(const char *base_dir, const char *dump_dir_name);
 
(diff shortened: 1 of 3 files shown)


jfilak:
Please add the declaration to internal_libreport.h and follow the common rules for exporting internal functions.

mhabrnal:
I updated the pull request and add testcase for `uid_in_group()`.

Is necessary to add  the declaration into `internal_libreport.h` in case the declaration is placed in `dump_dir.h` and `dump_dir.h` in included in `internal_libreport.h`?

jfilak:
> Is necessary to add the declaration into internal_libreport.h in case the declaration is placed in dump_dir.h and dump_dir.h in included in internal_libreport.h?

If you can answer to the oblique question how is the function related with other functions in `dump_dir.h`, then there is no need to declare it in `internal_libreport.h` (I didn't explicitly asked you to move the definition from `dump_dir.c` to better place in the first comment but I also didn't tell you to keep it there).

PR #400, opened 2015-12-16
RHEL6: dd: make function uid_in_group() public

mhabrnal:
Related to rhbz#1256705


jfilak:
Do you plan to backport the tests too?

mhabrnal:
It was an intention to do not backport the test but it's better to have the test here. So, I added the test as well.

PR #401, opened 2016-01-04
Offer users to create private ticket in GUI

jfilak:
This pull request adds support for letting users know that they can ask libreport to open a report visible only to specific people.


mhabrnal:
It works for me, I tested the feature even on MantisBT. We have to just fix the build (some tests are failing).

jfilak:
I fixed the failing create_private_ticket test (it was a bug in the test). The second failing test is cause by out-dated kernel on the builder (rhel-6 does no support O_PATH file descriptors).

mhabrnal:
Looks good to me, thank you!

PR #403, opened 2016-01-18
rhtsupport: use problme report API to create description

mhabrnal:
Related to: rhbz#1261398

Related testcase: https://github.com/abrt/abrt/pull/1106


mhabrnal:
We need just addURL to knowledge base article which describes all dump dir's elements into support case description.

PR #404, opened 2016-01-18
RHEL6: rhtsupport: use problem report API to create description

mhabrnal:
Related to: rhbz#1261398

Related testcase: https://github.com/abrt/abrt/pull/1107


mhabrnal:
We need just add URL to knowledge base article which describes all dump dir's elements into support case description.

PR #414, opened 2016-03-15
Workflow usability fixes

jfilak:
- document what needs to be documented
- publish functions necessary to get all configured workflows
- get rid of the need to know the default path to directory with workflows


mhabrnal:
Looks good to me. Thanks

PR #415, opened 2016-03-15
makefile: define LANG in release target

mhabrnal:


jfilak:
Thank you. Read the following link and fix the pull request please:
https://www.gnu.org/software/make/manual/html_node/Execution.html

mhabrnal:
PR updated.

PR #416, opened 2016-03-16
mailx: stop creating dead.letter on mailx failures

jfilak:
SELinux does not like mailx creating the file and I do not see any
reason to create the file in a problem directory because the file
contains a copy of the email that could not be send.

Failures of EVENT=notify are discoverable in system logs and if you run
the reporter manually, you will see that mailx failed.


mhabrnal:
Looks good to me.

PR #420, opened 2016-03-24
xml parser: be more verbose in case xml file cannot be opened

mhabrnal:
In a case a file cannot be openned by function
load_event_description_from_file(), no error messages was printed.


jfilak:
Thank you!

PR #422, opened 2016-03-30
RHEL7: Refined workflow

mhabrnal:
Do not open customers cases even if they do not require Red Hat Support to work on their cases.

Related to rhbz#1258482


jfilak:
Pleass add small marging between the Reproducible dropdown and the user hint.

mhabrnal:
Margin added. PR updated

jfilak:
All tests passed.

PR #427, opened 2016-04-05
RHEL7: lib: allow report SELinux denial from sealert under common user

mhabrnal:
The main purpose of the removed lines in this commit were preventing from
creating non-root dump dir's sub-directrories in the case an uid element
doesn't exist in time of creating the dump dir.

The removed lines are moved to the function problem_data_save() in abrt
src/lib/hooklib.c.

Related to rhbz#1264921


jfilak:
Thank you!

PR #428, opened 2016-04-07
Reduce false positive cases in ignored_words.conf

mhabrnal:
Related to rhbz#1319779
Related to rhbz#1311663


jfilak:
Please, squash these two commits into one.

mhabrnal:
updated

PR #430, opened 2016-04-07
RHEL7: use problem report api to define an emais' content

mhabrnal:
Related to rhbz#1281312


mhabrnal:
Related test case https://github.com/abrt/abrt/pull/1128

PR #433, opened 2016-04-19
dd: introduce functions getting occurrence stamps

jfilak:


mhabrnal:
looks good to me, test passed.

PR #437, opened 2016-05-23
dd: do not log in case the uid file doesn't exist

mhabrnal:
While journal crashes are extracted we try to read a uid file in time it
doesn't exist. The file is created later within dd_create_basic_files().
Adding DD_FAIL_QUIETLY_ENOENT prevents logging in case the file doesn't exist.


jfilak:
Thank you!

PR #441, opened 2016-07-08
wizard: do not create reproducible if complex_detail == no

### LINKED ISSUE #440: libreport-gtk always creates 'reporducible' element
The element is included in Bugzilla bug report description and it might confuse readers as the element always hold "Not sure how to reproduce the problem" message.

Guilty line: https://github.com/abrt/libreport/blob/master/src/gui-wizard-gtk/wizard.c#L2837

mhabrnal:
Fixes #440


jfilak:
Thank  you!

PR #448, opened 2016-08-25
bugzilla: stop including package details

jfilak:
The latest version of ABRT generate pkg_vendor and pkg_fingerprint
problem elements in dump directories. These elements have only a little
use in Bugzilla bug description.


mhabrnal:
We should exclude pkg_vendor and pkg_fingerprint from MantisBT bug description as well.

mhabrnal:
Excluding the files from MantisBT description is coveded by issue https://github.com/abrt/libreport/issues/457

PR #449, opened 2016-09-05
rhtsupport: fix a double free of config at exit

jfilak:
Introduced in commit 028b35b where I forgot on the code added in commit
5ff7f36.

Related: rhbz#1373094


mhabrnal:
retest this please

PR #461, opened 2016-11-01
test: yield error when you want to convert -1 to unsigned int

xsuchy:
addressing:
lt-map_string_get_set_as_various_types: libreport_types.at:193: main: Assertion `!try_get_map_string_item_as_uint(map, key, &retval) || !"Converts '-1' to unsigned number"' failed.
/builddir/build/BUILD/libreport-2.8.0.35.g20e5/tests/testsuite.dir/at-groups/36/test-source: line 262:  4688 Aborted                 (core dumped) $PRE_AT_CHECK ./map_string_get_set_as_various_types
stdout:
./libreport_types.at:44: exit code was 134, expected 0
36. libreport_types.at:44: 36. map_string_get_set_as_various_types (libreport_types.at:44): FAILED (libreport_types.at:44)

### DIFF
--- a/src/lib/abrt_types.c
+++ b/src/lib/abrt_types.c
@@ -131,8 +131,26 @@ int try_get_map_string_item_as_uint(map_string_t *ms, const char *key, unsigned 
     GET_ITEM_OR_RETURN(option, ms, key);
 
     char *endptr = NULL;
+
+    /* Check just negative number, positive ranges will be tested later */
+    long raw_signed_value = strtol(option, &endptr, 10);
+    if (raw_signed_value < 0)
+    {
+        log_warning("Value of option '%s' is out of unsigned integer range (bellow zero)", key);
+        return 0;
+    }
+
     errno = 0;
-    unsigned long raw_value = strtoul(option, &endptr, 10);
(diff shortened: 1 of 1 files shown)


xsuchy:
Pure strtoul() cannot be used because it happily convert negative number (and it is documented as feature).

jfilak:
Nice patch!

Can you please update the commit message to:
```
lib: refuse to parse negative number as unsigned int

man 3 strtoul:
Negative values are considered valid input and are silently converted 
to the equivalent unsigned long int value.
```

xsuchy:
Updated.

jfilak:
Do we really need to parse the string twice? If the number wasn't negative and wasn't out of range, we can use the return value of `strtol`. Or we can keep using only `strtoul` and upon a successful call we can check if there is the '-' prefix (don't forget to ignore leading white spaces) in the number string.

xsuchy:
Better?

jfilak:
I pushed a slightly modified commit:
https://github.com/abrt/libreport/commit/aff3bd331f866df2c8c5c6dc777ca1062ef3356b

Issue #463, opened 2016-11-07
reporter-s-journal: reporter should read SYSLOG_IDENTIFIER from env var


mhabrnal:

xsuchy:
@mhabrnal so is this done? Or not?

mhabrnal:
@xsuchy this issue is done.

PR #466, opened 2016-11-11
reporter-systemd-journal improvements

mhabrnal:
Related: #463

### DIFF
--- a/src/plugins/reporter-systemd-journal.c
+++ b/src/plugins/reporter-systemd-journal.c
@@ -308,7 +308,7 @@ int main(int argc, char **argv)
         problem_data_add_text_noteditable(problem_data, "crash_function", "??");
 
     /* Add SYSLOG_IDENTIFIER into problem data */
-    if (syslog_id)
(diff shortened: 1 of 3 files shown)


src/plugins/reporter-systemd-journal.c
```diff
@@ -308,7 +308,7 @@ int main(int argc, char **argv)
         problem_data_add_text_noteditable(problem_data, "crash_function", "??");
 
     /* Add SYSLOG_IDENTIFIER into problem data */
-    if (syslog_id)
+    if (syslog_id || (syslog_id = getenv("reporter_journal_syslog_id")))
```
jfilak: Is there any reason to use lowercase name? I think all libreport tools accept uppercase environment variables.

mhabrnal: updated

jfilak:
The PR looks good to me. Though the `run_event_on_problem_dir` function is not on the level I'd like to see, but let's roll out new version.

PR #467, opened 2016-11-18
client: add support for $releasever to debuginfo

jfilak:
ABRT needs to overwrite $releasever when analyzing a problem that
occurred in a container or a changed root environment.


mhabrnal:
retest this please

PR #468, opened 2016-11-23
introduce 'stop_on_not_reportable' option

jfilak:
Make it possible to ignore existence of not-reportable file and continue
in reporting.

The new configuration option is not yet persistent and can be configured
via an environment variable.

Related to #1351297
Related to abrt/abrt#1166


mhabrnal:
Looks good to me

Issue #459, opened 2016-10-31
[RFE] src/plugins/bugzilla.conf should default to value from /etc/os-release


jcpunk:
Right now src/plugins/bugzilla.conf is hard coded to use bugzilla.redhat.com.  However, Fedora 20+ and RHEL7 include a bugzilla url in /etc/os-release.

Defaulting to the value set in /etc/os-release allows for easier integration by respins.

jfilak:
Thank you for the issue. It make sense because libreport is already using `/etc/os-release` to get Bugzilla Product and Bugzilla Product Version.

PR #469, opened 2016-11-28
bugzilla use /etc/os-release for url

jcpunk:
My initial stab at #459 

I'm fairly sure that line 380 should test the value of os_release_map, but I've no idea how to check and see if it is empty.....

This should still allow the conf and the ENV to override the value, but provide an initial value for the setting.


trapas:

src/plugins/reporter-bugzilla.c
```diff
@@ -314,6 +314,20 @@ int main(int argc, char **argv)
 
     map_string_t *settings = new_map_string();
 
+    /* load some defaults from /etc/os-release */
```
jfilak: I the light of my recent container experience, I suggest you to first load the copy of the `/etc/os-release` file from the process' root that is stored in the problem directory under the name `os_info_in_rootdir`. Only if this first attempt to get the default Bugzilla URL for any reason (no problem directory, no such file, ...), fall back to loading of the regular `/etc/os-relase`.

Why I suggest this? Because the crashed processed might was in running in a container that were running a completely different distribution or an other version of the host distribution.

How to do that? The removals you have already done are OK and please keep them. However the additions should be amended with the following line and some more:
```
problem_data_get_osinfo(problem_data, osinfo);
```

You can find more details how to load `problem_data` here:
https://github.com/abrt/libreport/blob/master/src/plugins/reporter-bugzilla.c#L523

and how to use it here:
https://github.com/abrt/libreport/blob/master/src/plugins/reporter-bugzilla.c#L537

The `problem_data_get_osinfo` function takes care about the order of loaded `os-release` files:
https://github.com/abrt/libreport/blob/master/src/lib/problem_data.c#L640

jcpunk:
I will start looking at what needs to be done there.

I've hit one initial query, it looks like main() is building the problem data in your second link.  Could/Should this be moved earlier so that it isn't called twice?  What would be the best way to clean up afterward if the report isn't filed?

jcpunk:
I took a stab at the simplest solution I could see.  I fear I don't know enough C to have a good idea how to make the actual file name concatenation.

jcpunk:
After looking a bit more, I think I see the plan you suggested.  I'll give
it a try tomorrow and see if I can code it up.

jcpunk:
I think this is closer to what you were looking for.  It also has a hook so that other plugins could start using this behavior easily as well.

Once this looks viable overall, I can flatten all the changes into a single commit.

But first it needs to pass a review.....

src/plugins/reporter-bugzilla.c
```diff
@@ -312,21 +312,21 @@ int main(int argc, char **argv)
...
-    if (os_release != NULL)
+    /* pull in some defaults from os-release */
     {
-        map_string_t *os_release_map = new_map_string();
-        parse_osinfo(os_release, os_release_map);
-        log_debug("Loaded '/etc/os-release'");
-        bug_url = xstrdup(get_map_string_item_or_NULL(os_release_map, "BUG_REPORT_URL"));
-        if (bug_url != NULL)
-            rhbz->b_bugzilla_url = bug_url;
+        problem_data_t *problem_data = create_problem_data_for_reporting(dump_dir_name);
```
jfilak: This approach make sense to me. I have just two comments:
1. dump_dir_name might be NULL here
2. there is no need to call `create_problem_data_for_reporting` twice time

src/plugins/reporter-bugzilla.c
```diff
@@ -312,21 +312,21 @@ int main(int argc, char **argv)
...
+        if (!problem_data)
+            xfunc_die(); /* create_problem_data_for_reporting already emitted error msg */
+        else
+        {
+            map_string_t *osinfo = new_map_string();
+            problem_data_get_osinfo(problem_data, osinfo);
```
jfilak: Here you can also keep the osinfo variable populated for the future use.

src/plugins/reporter-bugzilla.c
jcpunk: I can wrap the first bit in an 'if', but if dump_dir is NULL, then osinfo wont get populated.

Is the best way to check if osinfo has data https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-get-keys and if it is empty it run problem_data_get_osinfo()?

jcpunk:
It looked like osinfo was being used to set a best effort default, so I decided on a function for best effort defaults that preps 'settings' so that it can override values cleanly and puts them next to each other for long term maintenance.  Rather than catch later on if we've no value, just set it and replace as needed.

In theory this should make setting other defaults that can be overridden easier and consistent....

Still not too sure what to do with the stuff around line 570 for if dump_dir_name is NULL.  On a related note, how would dump_dir_name be set to NULL?  It looks like it defaults to '.' with a command line override option.  Is there a way to set it to NULL via the command line (I tried reading lib/parse_options.c and got lost....)

I'm mostly sure   free_map_string(osinfo);  goes somewhere other than where I've got it....

jfilak:
>  In theory this should make setting other defaults that can be overridden easier and consistently....

This is interesting idea, however the current approach has been used because of collocation - the stuff that logically belongs together is stored together. When you need to add a new configuration option, you need to edit less functions.

> On a related note, how would dump_dir_name be set to NULL?

Good point. It cannot be NULL. I was under the impression that if you ommit -d on command line, dump_dir_name will be NULL. So you can assume that dump_dir_name is never NULL and its value points to a valid dump directory.

jcpunk:
The way I've got it right now `set_default_settings` is directly next to `set_settings` so that the behavior is basically adjacent.  I can inline it back under `main()`, but that felt less transparent as I was seeding the default content of the settings.  If keys change under the settings hash it seemed less likely that it would be seen down there.

On the `osinfo` front, I think I was able to shift all the calls to that variable together so that it can be free()d right away and the "no product" error block becomes a bit cleaner.

I'm thinking this may have drifted enough from my initial code that a squash to one commit might be worthwhile to clarify what all is in here at this point.  Thoughts?

jcpunk:
I've squashed it down into a single commit so that it should be a little easier to review going forward.

src/plugins/reporter-bugzilla.c
```diff
@@ -520,28 +554,11 @@ int main(int argc, char **argv)
...
-    problem_data_t *problem_data = create_problem_data_for_reporting(dump_dir_name);
-    if (!problem_data)
-        xfunc_die(); /* create_problem_data_for_reporting already emitted error msg */
-
     const char *component = problem_data_get_content_or_die(problem_data, FILENAME_COMPONENT);
     const char *duphash   = problem_data_get_content_or_NULL(problem_data, FILENAME_DUPHASH);
 //COMPAT, remove after 2.1 release
     if (!duphash) duphash = problem_data_get_content_or_die(problem_data, "global_uuid");
 
-    if (!rhbz.b_product || !*rhbz.b_product) /* if not overridden or empty... */
```
jfilak: I think you shouldn't have removed this `if` completely. We still need `repoter-bugzilla` to check that Product is configured properly.

jcpunk: I thought that was caught by line 445, I'll put the if back

jcpunk:
I believe I've got the commit message adjusted.

As I understand it the current commit:
- looks in os-release (as found by abrt)
- finds Product, Version, and BugURL
- sets the values in the `settings` hash
- does some basic error checking

Then, the code that reads in the `.conf` file stores any located values in the `settings` hash.  Possibly overwriting the values set by the new code.  The new code's manipulation of `settings` is physically next to the existing code's `settings` work for ease of maintenance.

Then the code checks the environment for any values to override the values from `.conf` and if needed overwrites any entries in `settings`.

src/plugins/reporter-bugzilla.c
jfilak: Aha, you are right, I didn't notice that line. However, you should not check Product configuration there because for example the command line argument '-t' does not require Product. The test should stay on line 562.

jcpunk: Done, I've removed the check on line 445 and left the check on 562 in place

jcpunk:
I think I've applied requested fixes and published a new squashed commit.

src/plugins/reporter-bugzilla.c
```diff
@@ -313,6 +331,19 @@ int main(int argc, char **argv)
...
+
+    /* pull in some defaults from os-release */
+    problem_data = create_problem_data_for_reporting(dump_dir_name);
+    if (!problem_data)
+        xfunc_die(); /* create_problem_data_for_reporting already emitted error msg */
+    else
+    {
+        map_string_t *osinfo = new_map_string();
+        problem_data_get_osinfo(problem_data, osinfo);
+        set_default_settings(osinfo, &settings);
```
jfilak: The 2nd parameter of the `set_default_settings` function is `map_string_t *` but you are passing `map_string_t **` here.

src/plugins/reporter-bugzilla.c
```diff
@@ -86,6 +86,24 @@ struct bugzilla_struct {
...
+static void set_default_settings(map_string_t *osinfo, map_string_t *settings)
+{
+    char *default_BugzillaURL;
+    parse_osinfo_for_bug_url(osinfo, &default_BugzillaURL);
+    /* if BugzillaURL is defined in conf_file or env , it will replace this value */
+    set_map_string_item_from_string(settings, "BugzillaURL", default_BugzillaURL);
+    log_debug("Loaded BUG_REPORT_URL '%s' from os-release", default_BugzillaURL);
+
+    char *default_Product;
+    char *default_ProductVersion;
```
jfilak: Who will free() these variables?

src/lib/parse_release.c
```diff
@@ -289,6 +289,19 @@ void parse_osinfo_for_bz(map_string_t *osinfo, char** product, char** version)
...
+
+    if (os_bug_report_url)
+        *url = xstrdup(os_bug_report_url);
+    else
+        /* unset or something bad happend */
+        *url = NULL;
+
+    return;
+}
+
```
jfilak: Do we really need this function? What is the benefit? It reads just one key from a single map and there is no magic hidden there.

jfilak:
Yes, the pull request looks good to me. There some minor issues but will fix them. It would be nice if you could create a test for this PR. The test should live in github.com/abrt/abrt and here you can find how to create it: https://github.com/abrt/abrt/blob/master/HACKING.md#verify-abrt-plays-nice-with-the-rest-of-os

I suggest to create a new test called "bugzilla-default-url".

src/lib/parse_release.c
jcpunk: Mostly it was an attempt to follow the behavior of the other "parse_osinfo_for" functions.  In particular it was inspired by the `parse_osinfo_for_bz` function.  Long term, I'm planning to swing back around to the other reporter plugins and add similar "default to using `BUG_REPORT_URL`" logic over time.  I was hopeful that a somewhat central function might make future patches less complex and easy to merge.

I could just in line it over in the bugzilla reporter, but this just sorta "felt right" over here since I was planning to do it again and again...

src/plugins/reporter-bugzilla.c
jcpunk: oh dear, I'll get a fix pushed up ASAP.  I'll hold off squashing for now due to conversation on `parse_release.c`

src/plugins/reporter-bugzilla.c
jcpunk: Good question.... my lack of recent `C` development is showing... my brain said "It went out of scope so don't worry."

I'll add the `free()` calls how I think they should look.

I'm not totally sure I'm calling `parse_osinfo_for_bz` with the right kind of arguments based on what you said regarding `set_default_settings` below.  I've been working in higher level languages for a while and it seems my familiarity with pointers is less good than I thought...

src/lib/parse_release.c
jfilak: Alright then. Since this is a public function we should cover it with unit tests. Please add a new test case to https://github.com/abrt/libreport/blob/master/tests/osinfo.at . Here you can find how the unit tests work: https://github.com/abrt/abrt/blob/master/HACKING.md#how-to-check-your-changes.

Just a side note, the function can be used in `reporter-mantisbt` too.

jfilak:
Great! Thank you for the fixes. Once we finish the tests we can merge this PR.

jcpunk:
Following the logic I found in `libreport/tests/osinfo.at` I think I've got the tests in place for the public api function.

It seemed like it belonged here rather than in abrt.

jcpunk:
I do mean this request seriously.  I'm able to compile it in my Fedora VM.

jcpunk:
I can't seem to find what is wrong with my test fprintf so I can see why it
is failing the test.

It is becoming more obvious that I've forgotten all my C.....

jfilak:
OK, let me finish the pull request then. It will take me 2-3 days as I can work on ABRT matters only ~1 hour a day.

jcpunk:
Thanks!  That is probably for the best.

PR #471, opened 2016-12-13
configure.ac: Remove nss dependency

wlindauer:
This does not seem to be referenced anywhere.

I might be missing something, but I cannot find where this is used. If it is not needed, can you please merge? If it is still needed, I am sorry for the churn and would like to know what it is used for.


trapas:

jfilak:
The NSS requirement is a left-over from the times where libreport and abrt were a single project. Thank you for spotting this.

PR #480, opened 2017-03-14
reporter-s-journal: add journal entry PROBLEM_DIR

mhabrnal:
The PROBLEM_DIR contains path to problem directory.
This features is requested by ABRT cockpit plugin.

### DIFF
--- a/src/plugins/reporter-systemd-journal.c
+++ b/src/plugins/reporter-systemd-journal.c
@@ -102,6 +102,7 @@ static void msg_content_free(msg_content_t *msg_c)
 #define BINARY_NAME "binary"
 #define SYSLOG_ID   "SYSLOG_IDENTIFIER"
 #define MESSAGE_ID  "MESSAGE_ID"
+#define DUMPDIR_PATH "DIR"
 
 enum {
     DUMP_NONE      = 1 << 0,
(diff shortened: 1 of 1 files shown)


marusak:
I asked @mhabrnal to implement this, since it was very painful to connect journal logs to abrt problems. I tested it and it works as expected. Thanks:)

jfilak:
@mhabrnal I thought the deal was to save UUID and DUPHASH because the problem directory might be removed at the time of reading journald but a dupe can be present there and the dupe can be identified based on these two values.

mhabrnal:
@jfilak Adding UUID as an another journal entry is a good idea. DUPHASH  element cannot be present in problem dir in the time when abrt-action-notify is executed because it's created by abrt-action-analyze-backtrace which is not run automatically.

jfilak:
That's true only for CCpp problems.

marusak:
@jfilak So what is the best identifier? Is it UUID? Is UUID unique and always available? @mhabrnal updated this PR so it contains also UUID and DUPHASH. Is it now ok?

jfilak:
Try to get a Problems2 entry using all these identifiers in this order -> DUMPDIR_PATH, DUPHASH, UUID.

Thank you for the notice! We could merge this PR if the commit message were correct (PROBLEM_UUID ?) and and the commit included an explanation why we need to use absolute file system path instead of a directory name.

marusak:
Thanks, I will to that.

Well, then somebody will have to fix that, since @mhabrnal will not be in work for some time.
I will take a look on it tomorrow or if you have some time, feel free to do that.

marusak:
@jfilak updated the commit message. Is it better now?

jfilak:
Excellent job!

jfilak:
Please, add an entry describing this PR to CHANGELOG.md.

PR #481, opened 2017-03-21
reporter-s-journal: add count to default logs

marusak:
Field count is available only in FULL dump. There are however two good
reasons why to include this in NONE dump.

1. It is quite useful to know if this problem occurred first time or is it
problem that happens a lot.

2. We need it for cockpit integration. If the problem happened more times it
will have multiple logs. By comparing count in the log and real count, we can
know, if this is the last log of the problem. (timestamps do not work ideally
for this, since the time of problem does not equal to the time of the
log - there is some delay)

Signed-off-by: Matej Marusak <mmarusak@redhat.com>

### DIFF
--- a/src/plugins/reporter-systemd-journal.c
+++ b/src/plugins/reporter-systemd-journal.c
@@ -126,6 +126,7 @@ static const char *const fields_default[] = {
     DUMPDIR_PATH                ,
     FILENAME_UUID               ,
     FILENAME_DUPHASH            ,
+    FILENAME_COUNT              ,
     NULL
 };
 


trapas:

marusak:
test this please

jfilak:
Out of curiosity, why do you need to know if a journald entry is the last one?

marusak:
Because I don't what to show all occurrences of the problem.
Imagine this is you journal (from newest to oldest with its count):
- problem_1 3
- problem_1 2
- problem_2 2
- problem_2 1
- problem_1 1
And what I really want to display is:
- problem_1 3
- problem_2 2

jfilak:
Can you please resolve the conflict?

marusak:
Done

jfilak:
Thank you!

jfilak:
Please, add an entry describing this PR to CHANGELOG.md.

PR #483, opened 2017-04-07
lib: fix newline issue with ask_password

mkutlak:
The newline wasn't printed out because printing characters in terminal was
disabled (ECHO flag unset). After 'ask_password' function call, there would
have to be print function to create newline, otherwise the next message
would show up in the same line.

ECHONL flag allows to print newline character even if the ECHO flag is disabled.

Logic for the changes of the flags follows:
Change ECHO flag only when:
  - ECHO is set (0) and there was no change (0) previously.
  OR
  - ECHO is unset (1) and there was a change (1) previously.

Change ECHONL flag only when:
  - ECHONL is unset (1) and there was no change (0) previously.
  OR
  - ECHONL is set (0) and there was a change (1) previously.

Also ECHONL needs ICANON flag to be set (It is set by default).


jfilak:
Thanks. Good job! Have you considered defining a test for this change?

mkutlak:
Sure, I will make one.

jfilak:
Are you going to fix the build failure? Or is the failure unrelated to your changes?

mkutlak:
The build failure is from test I created. I am trying to fix the failures.

mkutlak:
I removed the test from the commit as it was unusable. 

The problem with the test was that it needed STDIN_FILENO to refer to a terminal, which it didn't.

jfilak:
Hmm, what about to run the test in a new process started by `forkpty`.
```
pid_t pid = forkpty(...);
if (pid < 0) {
    perr_msg_and_die("fork");
}

if (pid == 0) {

    /*
        Add your test code here .
     */

    return EXIT_SUCCESSFUL;
}

int status;
if (safe_waitpid(pid, &status, 0) < 0) {
    perr_msg_and_die("waitpid");
}

if (WIFEXITED(status)) {
    return WEXITSTATU(status);
}
else if (WIFSIGNALED(status)) {
    err_msg_and_die("killed by signal %d", WTERMSIG(wstatus));
}
else {
     err_msg_and_die("test unexpectedly stopped/continued");
}
```

jfilak:
Well done! Thank you!

mkutlak:
Thanks to you too. You helped me a lot.

jfilak:
Please, add an entry describing this PR to CHANGELOG.md

PR #485, opened 2017-04-25
lib: replace hash table with list

mkutlak:
Order of login credentials wasn't guaranteed, because event options with missing
values were inserted into hash table and then extracted from it.

I renamed function validate_event as its name was a little bit confusing. The function
returns options that are missing input values and error message should be displayed.

This fixes abrt/abrt#1231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ in libreport configs.
 - Update glib minimal version dependency, this bug caused update problems when
 updating libreport and pre-2.28 glib was installed.
 - Fix missing newline when asking for password.
+- Fix a bug causing inconsistent order of username and password fields when
+reporting bugs.
 
 
 ## [2.9.1] - 2017-03-16
--- a/src/include/event_config.h
+++ b/src/include/event_config.h
@@ -63,6 +63,17 @@ typedef struct
     bool is_advanced;
 } event_option_t;
 
+/*
+ * struct holds
+ *   invopt_name = name of the option with invalid value
+ *   invopt_error = string of the error message
+ */
+typedef struct
+{
+    char *invopt_name;
+    char *invopt_error;
+} invalid_option_t;
+
 event_option_t *new_event_option(void);
 void free_event_option(event_option_t *p);
 
@@ -108,6 +119,8 @@ bool ec_restricted_access_enabled(event_config_t *ec);
 
 void free_event_config(event_config_t *p);
 
+invalid_option_t *new_invalid_option(void);
+void free_invalid_options(invalid_option_t* p);
 
 void load_event_description_from_file(event_config_t *event_config, const char* filename);
 
@@ -126,7 +139,7 @@ extern GHashTable *g_event_config_list;   // for iterating through entire list o
 GList *export_event_config(const char *event_name);
 void unexport_event_config(GList *env_list);
 
-GHashTable *validate_event(const char *event_name);
+GList *get_options_with_err_msg(const char *event_name);
 
 /*
  * Checks usability of problem's backtrace rating against required rating level
--- a/src/lib/event_config.c
+++ b/src/lib/event_config.c
@@ -22,6 +22,11 @@
 GHashTable *g_event_config_list;
 static GHashTable *g_event_config_symlinks;
 
+invalid_option_t *new_invalid_option(void)
+{
+    return xzalloc(sizeof(invalid_option_t));
+}
+
 event_option_t *new_event_option(void)
 {
     return xzalloc(sizeof(event_option_t));
@@ -119,6 +124,15 @@ bool ec_restricted_access_enabled(event_config_t *ec)
     }
 
     return eo->eo_value != NULL && string_to_bool(eo->eo_value);
+}
+
+void free_invalid_options(invalid_option_t *p)
+{
+    if (!p)
+        return;
+    free(p->invopt_name);
+    free(p->invopt_error);
+    free(p);
 }
 
 void free_event_option(event_option_t *p)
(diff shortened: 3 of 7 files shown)


tests/event_config.at
```diff
@@ -58,3 +58,130 @@ TS_MAIN
 }
 TS_RETURN_MAIN
 ]])
+
+## ------------------------ ##
+## get_options_with_err_msg ##
+## ------------------------ ##
+
+AT_TESTFUN([get_options_with_err_msg],
+[[
```
jfilak: Please join these two lines above, otherwise gcc errors/warnings (even runtime error messages) will be pointing to wrong line numbers.

tests/event_config.at
```diff
@@ -58,3 +58,130 @@ TS_MAIN
...
+#define EVENTS_CONF_FIR "../../conf/bugtester"
+
+static int cmp_event_option_name_with_string(gconstpointer a, gconstpointer b)
+{
+    const event_option_t *evopt = a;
+    return !evopt->eo_name || strcmp(evopt->eo_name, (char *)b) != 0;
+}
+
+
+static void load_config_files(const char *dir_path)
```
jfilak: Do you really need to load the files? What about to just create the option lists manually and expect the same content in the configuration files? You have already implemented the parsing, but this function is a bit complex and thus might contain bugs - I prefer static content in tests, as it is less error prone.

src/lib/event_config.c
```diff
@@ -427,29 +428,31 @@ static char *validate_event_option(event_option_t *opt)
...
+        }
     }
 
-    if (g_hash_table_size(errors))
-        return errors;
+    if (err_list != NULL)
+        return g_list_reverse(err_list);
 
-    g_hash_table_destroy(errors);
+    g_list_free(err_list);
```
jfilak: Well, err_list is NULL on this line.

src/include/event_config.h
```diff
@@ -126,7 +127,7 @@ extern GHashTable *g_event_config_list;   // for iterating through entire list o
 GList *export_event_config(const char *event_name);
 void unexport_event_config(GList *env_list);
 
-GHashTable *validate_event(const char *event_name);
+GList *get_options_with_err_msg(const char *event_name);
```
jfilak: I am sorry but I do not think this is a good idea. The function has an unexpected side effect. You cannot call it twice - if you do you get a memory leak - and the `eo_error_msg` member become nonsense when you set the value.

The function `validate_event` was supposed to return a list of options with invalid values (no value is an invalid value for mandatory options). The original author used `GHashTable` as the easiest way of passing the pairs. What about to create a new structure, let's say `struct invalid_options { const char *opt_name; char *error; };`, and return `GList *` of these structures.

src/report-newt/report-newt.c
```diff
@@ -153,17 +152,17 @@ static int configure_reporter(struct reporter *r, bool skip_if_valid)
...
 
             /* Catenate the error messages */
             buf[0] = '\0';
-            for (g_hash_table_iter_init(&iter, error_table);
-                    g_hash_table_iter_next(&iter, (void**)&opt_name, (void**)&err_msg); )
+            for (iter = error_table; iter; iter = iter->next)
             {
-                opt = get_event_option_from_list(opt_name, r->config->options);
+                event_option_t *op_data = (event_option_t *)iter->data;
+                opt = get_event_option_from_list(op_data->eo_name, r->config->options);
```
jfilak: One would guess that `op_data` and `opt` hold the same pointer here.

tests/event_config.at
mkutlak: You are right. It is not needed. I will replace the loading of the files with manually created event config with option list.

tests/event_config.at
jfilak: If you are not sure what I have in mind, please have a look at:
https://github.com/abrt/libreport/blob/master/tests/local.at#L22

The test case is compiled as a local file which is a concatenation of:
```
#line __line__ "__file__"
```
and the test contents.

So, if there is a warning, an error, or  a crash the source file is `event_config.at` with correct line number and the temporary compilation file.

https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html

jfilak: https://github.com/abrt/libreport/commit/4d9e6f7835a22c435756b229e00b50c52d996950

mkutlak:
I've updated the code to reflect your requested changes. Thank you for your feedback.

jfilak:
Sorry for the late notice but this PR definitely deserves a commit updating CHANGELOG.md.

src/include/event_config.h
```diff
@@ -63,6 +63,17 @@ typedef struct
...
 
+/*
+ * struct holds
+ *   opt_name = name of the option with invalid value
+ *   error = string of the error message
+ */
+typedef struct
+{
+    char *opt_name;
+    char *error;
```
jfilak: I need to correct myself and I must apologize for not reviewing my own comments (I wrote my comment after midnight, that may explain something).

When I joined ABRT team, I was taught that the prefix of structured members is good for greping. If you use "opt_", it will not be possible to distinguish between uses of `invalid_option_t` and `event_option_t`. The same apply for the member error - there are plenty of  `error` occurrences in libreport source codes.

I suggest using the prefix `invopt_` for members of the structure `invalid_option_t`.

mkutlak:
Done.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Data structure for invalid options in events was changed from hashtable to
+list.
+
```
jfilak: The main reason for this PR is the bug causing that libreport asks for 'password' before it asks for 'login'. How about to say this instead of ...

@juliusmilan can you please help him to write a more informative message - I'm not going to do that intentionally - I'm too lazy :)

mkutlak:
retest this please

PR #495, opened 2017-06-15
dd: extend create_dump_dir to allow set pid in dumpdir name

mhabrnal:


jfilak:
Thank you!

PR #496, opened 2017-06-29
bugzilla: add check if option -d was entered

mkutlak:
The change from commit 946c5 caused that reporter was ALWAYS trying to read
from problem directory. And because the default value for it is set to ".",
running reporter from outside of problem directory caused errors.


jfilak:
Thank you!

PR #498, opened 2017-08-23
anaconda: run log-capture if available

jcpunk:
Newer anaconda releases provide a tool to capture anaconda logs.  This may be handy for automatic reporting in order to understand why an install crashed.

The attached patch conditionally runs the script if it exists.  The first arg to log-capture is where to store the output.

### DIFF
--- a/src/plugins/bugzilla_anaconda_event.conf
+++ b/src/plugins/bugzilla_anaconda_event.conf
@@ -7,6 +7,9 @@ EVENT=report_Bugzilla component=anaconda
 			sed 's/^.*rootpw.*$/<auto-removed line containing rootpw>/' -i $sf
 		fi
 	done
+	if [ -x /usr/libexec/anaconda/log-capture ]; then
+		/usr/libexec/anaconda/log-capture "$DUMP_DIR/log-capture.tar.bz2"
+	fi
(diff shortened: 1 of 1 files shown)


trapas:

mkutlak:
ok to test

jfilak:
good job! thanks!

jfilak:
@jcpunk should we allow users to review the logs? And perhaps remove sensitive information?

jcpunk:
> should we allow users to review the logs? And perhaps remove sensitive information?

That is probably advisable.

In theory the sensitive stuff has been scrubbed.  But in practice, lots of folks have differing ideas about what is sensitive.  Giving them the option is probably for the best.

jfilak:
@jcpunk Then we cannot attach the logs in an archive and we have to let libreport know that it should offer this anaconda log for review - i.e. we have to add that file to the list of editable files at:
https://github.com/abrt/libreport/blob/master/src/lib/problem_data.c#L267

Well, how does python-meh deal with anaconda crashes?

Hmm, mentioning python-meh, I was under the impression that anaconda reports contain all the necessary information - including this log.

@M4rtinK what do you think about this pull request?

PR #499, opened 2017-08-24
logging: rename log() to log_warning()

mhabrnal:
Renaming log() logging function to log_warning() because
of conflict with log() function from math.h.

Related to #1481205


juliusmilan:
lgtm

PR #503, opened 2017-09-15
Adding unpackaged dump dir element

mhabrnal:


src/include/internal_libreport.h
```diff
@@ -1148,6 +1148,7 @@ struct dump_dir *open_directory_for_writing(
 #define FILENAME_PKG_VERSION   "pkg_version"
 #define FILENAME_PKG_RELEASE   "pkg_release"
 #define FILENAME_PKG_ARCH      "pkg_arch"
+#define FILENAME_UNPACKAGED    "unpackaged"
```
jfilak: Why do you need the new file? What about to call a report "unpackaged" if the file `package` does not exist?

src/plugins/ureport.conf
```diff
@@ -36,3 +36,6 @@
 # HTTPAuth = rhts-credentials
 # Use username and password:
 # HTTPAuth = username:password
+
+# Processing problems coming from unpackaged executables
+ProcessUnpackaged = no
```
jfilak: I would rather introduce this option in the ABRT's ureport-reporter wrapper `abrt-action-ureport`.

xsuchy: Can you elaborate why?

We decided for this config because there is the setting URL. So here you specify where to send the ureports and what to send.

jfilak: Sorry for being so quiet. I would rather see this option in ABRT because I consider `reporter-ureport` as a low level, universal tool that just executes action it is asked for. That means that the caller should check if he wants to send a report for unpackaged stuff. I believe that other report programs work this way too - reporter-bugzilla, reporter-mantisbt, reporter-logger, ... If I am wrong, I take this comment back.

You are right with the configuration options, that makes my point weaker :) Interesting enough, `abrt-action-ureport` parses this configuration file:
https://github.com/abrt/abrt/blob/master/src/plugins/abrt-action-ureport

Botton line, my comment expresses my feeling - if you disagree, then you can ignore the comment.

jfilak: The option must be introduced in the event configuration file too (to allow users to change this configuraiton option from GUI):
https://github.com/abrt/libreport/blob/master/src/plugins/report_uReport.xml.in

That's how other reporters work.

For example, the EVENT configuration option `uReport_WatchReportedBugs` is recognized only by the utility `abrt-action-ureport`. The reporter knows how to start watching repoted bugs but it is another reporting step, so a high level tool should ask the reporter to do that once the first step is completed (FAF accepted the ureport).

PR #506, opened 2017-09-20
remove old changelogs

xsuchy:


jfilak:
Changelog should keep less than 2 years old entries only? I am asking because I want to lear how to do packaging correctly.

xsuchy:
the guideliness, keep just recent. It is up to the maintainer, what "recent" means. For me - 2 years sounds fair. :)

PR #508, opened 2017-09-20
client-python,report-python: Allow python to be optional at build time

wlindauer:
This allows python2 and python3 to be independently used for builds that
only require using one or the other.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,18 +56,6 @@ IT_PROG_INTLTOOL([0.35.0])
 
 dnl ****** END ****************************************
 
-AM_PATH_PYTHON
-if test -z "$PYTHON"; then
-    echo "*** Essential program python not found" 1>&2
-    exit 1
-fi
-
-AC_PATH_PROG([PYTHON3], [python3], [no])
-if test -z "$PYTHON3"; then
-    echo "*** Essential program python3 not found" 1>&2
-    exit 1
-fi
-
 AC_PATH_PROG([ASCIIDOC], [asciidoc], [no])
 [if test "$ASCIIDOC" = "no"]
 [then]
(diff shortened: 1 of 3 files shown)


trapas:

jfilak:
ok to test

jfilak:
@mhabrnal  @marusak  @mkutlak  @xsuchy What causes the failing tests?

mkutlak:
The failed test was caused by an unrelated issue on the testing server. 
@trapas retest this please

mkutlak:
Ignore my previous comment, I thought this was different PR. 

The test failed because of this:
*** No rule to make target '__init__.py', needed by 'all-am'.  Stop.

jfilak:
@mkutlak Which directory? What about to make the build log public - upload it to your fedorapeople space, please.

mkutlak:
Found the issue:
https://www.gnu.org/software/automake/manual/automake.html#Uniform

@wlindauer Change the _DATA suffixes back to _PYTHON and it should work.

wlindauer:
@mkutlak Thanks for looking into it. I can resubmit with those changes. The reason I had changed the _PYTHON to _DATA is that I was under the impression that _PYTHON only byte-compiled into python2 output and was not appropriate for python3. But I can't really find documentation stating that. Do you know if this is true or not?

wlindauer:
@mkutlak I have pushed an updated commit with the new changes. Thanks for your help.

mkutlak:
@wlindauer Sorry for not responding. 
I am not that experienced with automake, so I don't know the answer to your question. Perhaps @jfilak would know more.

jfilak:
Thank you!

PR #517, opened 2018-01-02
reporter-{bugzilla,mantisbt,rhtsupport}: fix free

juliusmilan:
There was an incorrect free of string local_conf right after addition to
the list conf_file, which caused that presence of the pointed string in
the memory was undeterministic. As a result, local config was sometimes
used and sometimes not.


mkutlak:
Thank you!

PR #524, opened 2018-04-09
Move uReport workflow to ureport plugin

jcpunk:
This moves the ureport workflow over to the ureport-plugin so that it isn't dependent on the rht-support plugin.

### DIFF
--- a/libreport.spec.in
+++ b/libreport.spec.in
@@ -307,6 +307,7 @@ install this package and you're done.
 %package rhel
 Summary: Default configuration for reporting bugs via Red Hat infrastructure
 Requires: %{name} = %{version}-%{release}
+Requires: %{name}-plugin-ureport
 
 %description rhel
 Default configuration for reporting bugs via Red Hat infrastructure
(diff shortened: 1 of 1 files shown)


trapas:

xsuchy:
test this please

mkutlak:
RPM build errors:
error: File not found: /builddir/build/BUILDROOT/libreport-2.9.3.34.g9cf1-1.fc28.x86_64/etc/libreport/workflows.d/report_uReport.conf
error: File not found: /builddir/build/BUILDROOT/libreport-2.9.3.34.g9cf1-1.fc28.x86_64/usr/share/libreport/workflows/workflow_uReport.xml
error: File not found: /builddir/build/BUILDROOT/libreport-2.9.3.34.g9cf1-1.fc28.x86_64/usr/share/man/man5/report_uReport.conf.5.*

jcpunk:
Ahh, tested it on a EL7 box, should be fixed now for fedora as well.

mkutlak:
_Rebased + added commit that modifies the original commit._

The workflow_uReport is specific for RHEL and we don't want to ship it in Fedora.

Otherwise this would happend:
```
$ abrt report

1 Report to Fedora
  Process the python 3 exception using the Fedora infrastructure

2 Submit anonymous crash report
  Submit anonymous crash report - I do not want to be contacted by Red Hat Support

Select a workflow to run:
```

mkutlak:
@trapas add to whitelist

ernestask:
To be honest, I don’t see how the original premise holds, because I can’t see anything in the spec that depends on the rhtsupport plugin and has anything to do with the ureport plugin. Not to mention that the workflow is only useful in RHEL.

ernestask:
Squashed and merged as https://github.com/abrt/libreport/commit/54c08be9511c418b90c1ad8d1576294cf073fce7.

PR #540, opened 2018-11-15
gtk-helpers: config_dialog: Hide tree view header

ernestask:
There’s only one column and the model used is not sortable, so it
should just be hidden.

Fixes https://github.com/abrt/gnome-abrt/issues/150


mkutlak:
LGTM. Thank you.

PR #541, opened 2018-12-01
report-client: Find debuginfos in own method

marusak:
This commit picks a piece of code responsible for looking up debuginfo
packages into own method. For making sure that it works in any code that
used it the old way, it checks if the method for finding packages was
already called and if not, calls it directly.

Purpose of this change is to be able to get size of packages those are
going to be installed before installing them. In that case anyone using
these downloader classes is going to be able for example prepare enough
space for the newly installed packages.

Related to #811978.

Signed-off-by: Matej Marusak <mmarusak@redhat.com>


xsuchy:
+1

PR #542, opened 2018-12-03
Improve backtrace generation

mkutlak:
The current solution finds packages for given build-ids and then
downloads them. The problem is that some debuginfo packages require
other packages and if they are not available the generated backtrace
becomes unusable.

This commit adds a query for required packages and downloads them together
with the main packages.

And also modify the error message when the backtrace is unusable.

Related: rhbz#1515265


mkutlak:
@marusak I've added recursive function to find required packages for required packages.

PR #545, opened 2018-12-06
lib: Explicitly do not use DST

marusak:
DST = Daylight Saving Time
Value of this field is not set up by strptime and therefore value
of `tm_isdst` field is undefined. Bacause of this reason, output of
`mktime` may differ +-1 hour.

@mkutlak this should fix tests 26 and 97 on rawhide, can you verify please?


mkutlak:
Verified. :+1: 

All tests passed on my rawhide machine.

PR #546, opened 2018-12-21
Update dd_get_owner prototype

webera:
Update dd_get_owner prototype and implementation to properly handle the
possible error return values which can be negative.

As reference the link to the issue in abrt repo:
https://github.com/abrt/abrt/issues/1345

Basically `dd_get_owner()` calls `read_number_from_file_at()` which returns negative numbers in case of error: 

https://github.com/abrt/libreport/blob/5e0c2c3839f7205a90aedd6be7a1f54bc4306ebb/src/lib/dump_dir.c#L240

If this patch makes sense it must be ensured that the users off this function
are adapted.

### DIFF
--- a/src/include/dump_dir.h
+++ b/src/include/dump_dir.h
@@ -346,7 +346,7 @@ int dd_set_no_owner(struct dump_dir *dd);
  * If meta-data misses owner, returns fs owner.
  * Can be used with DD_OPEN_FD_ONLY.
  */
-uid_t dd_get_owner(struct dump_dir *dd);
+int dd_get_owner(struct dump_dir *dd, uid_t *owner);
 
 /* Returns UNIX time stamp of the first occurrence of the problem.
  *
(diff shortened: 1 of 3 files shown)


xsuchy:
@trapas test this please

src/include/dump_dir.h
```diff
@@ -346,7 +346,7 @@ int dd_set_no_owner(struct dump_dir *dd);
  * If meta-data misses owner, returns fs owner.
  * Can be used with DD_OPEN_FD_ONLY.
  */
-uid_t dd_get_owner(struct dump_dir *dd);
+int dd_get_owner(struct dump_dir *dd, uid_t *owner);
```
mkutlak: Wouldn't it be better to return the `uid_t` result and have some kind of `errorno` parameter? 
That way we could use the function directly and test the value of `errorno` for errors. And if we don't care about errors just pass it `NULL`.

xsuchy:
I am closing this as PR https://github.com/abrt/libreport/pull/570 will be merged. Thank you for your contribution.

PR #547, opened 2019-01-04
rhbz: Replace nomail flag with minor_update

mkutlak:
In Bugzilla v5.0 nomail flag is replaced with minor_update. [1]
Because we still use the old nomail flag emails are being sent for every added attachment/comment.

Related: rhbz#1660157
Closes: abrt/abrt#1346

See also:
- https://partner-bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/WebService/Bug.html?highlight=minor_update

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1655829


ernestask:
Makes sense to me.

PR #549, opened 2019-01-10
Export used workflow to reported_to

ernestask:
See commits.


src/plugins/reporter-rhtsupport.c
```diff
@@ -644,7 +648,10 @@ int main(int argc, char **argv)
                 int yes = ask_yes_no(msg);
                 free(msg);
                 if (!yes)
+                {
+                    g_free(url);
```
mkutlak: Freed `url` is used on the following line.

src/plugins/reporter-ureport.c
```diff
@@ -224,20 +232,23 @@ int main(int argc, char **argv)
...
+            if (!url)
                 error_msg_and_die(_("The report result '%s' is missing URL."), report_result_type);
 
             /* Avoid the need to duplicate the string. */
-            attach_value = attach_value_from_rt_data = result->url;
-            result->url = NULL;
-
-            free_report_result(result);
+            attach_value = url;
+            attach_value_from_rt_data = url;
```
mkutlak: You should free the `url`.

ernestask:
When testing abrt locally (and even building docs), the installed `abrt-dbus` will not run due to `libreport_free_report_result()` no longer being exported. One would need to run the modified version from source.

src/plugins/reporter-ureport.c
ernestask: If you take a look at the `finalize` label, `attach_value_from_rt_data` is freed, which now holds `url`. It’s annoyingly complicated, but I didn’t notice a path that results in a leak.

src/plugins/reporter-rhtsupport.c
ernestask: Ugh, every time.

ernestask:
Updated. The URL weirdness in gnome-abrt is caused by the fact that it does its own parsing of the `reported_to` file. PR for that coming as well.

tests/report_result.at
```diff
@@ -0,0 +1,125 @@
...
+    assert(NULL != strbuf || !"Only TIME");
+    assert(string_cmp("OnlyTIME: TIME=2000-01-01-00:00:00", strbuf->buf));
+
+    strbuf_free(strbuf);
+    report_result_free(result);
+
+    result = report_result_new_with_label("OnlyWORKFLOW");
+    report_result_set_workflow(result, "workflow_FedoraCCpp");
+    strbuf = report_result_to_string(result);
+    assert(NULL != strbuf || !"Only TIME");
```
mkutlak: `s/Only TIME/Only WORKFLOW/`

mkutlak:
You should bump the glib_version in spec file as the `g_auto` are in [glib since 2.44](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#g-auto).

https://github.com/abrt/libreport/blob/5566b41d94e34ea616ec70964f1c8f96cdc346f0/libreport.spec.in#L13

tests/report_result.at
ernestask: Oops.

ernestask:
> You should bump the glib_version in spec file as the `g_auto` functions are in [glib since 2.44](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#g-auto).

Indeed, thanks.

ernestask:
The docs will show only the first stable version that introduced the API, but the macros are there since 2.43.4.

PR #552, opened 2019-01-16
Enhance emergeny analyses output

marusak:
See https://bugzilla.redhat.com/show_bug.cgi?id=1540500 and commit messages.

I am more than happy to discuss the message that should be printed. This was my first idea. We need to inform user that when click the URL, is not going to be able to see the archive. The url looks like this: https://retrace.fedoraproject.org/faf/dumpdirs/new/oops-2018-01-26-21:28:20-1347-0.tar.gz/

### DIFF
--- a/src/plugins/emergencyanalysis_event.conf
+++ b/src/plugins/emergencyanalysis_event.conf
@@ -2,5 +2,6 @@ EVENT=report_EmergencyAnalysis
     rpm -q abrt > abrt_installed_version 2>/dev/null
     rpm -q libreport > libreport_version 2>/dev/null
     rpm -q satyr > satyr_version 2>/dev/null
+    echo -e "An archive will be created.\nThis archive is accessible only by ABRT developers.\nTherefore if you try to open the produced web location you will not be able to see it."
     EXCLUDE_FROM_REPORT="" reporter-upload 1>/dev/null
     rm -f abrt_installed_version libreport_version satyr_version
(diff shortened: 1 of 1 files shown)


src/plugins/emergencyanalysis_event.conf
```diff
@@ -2,5 +2,6 @@ EVENT=report_EmergencyAnalysis
     rpm -q abrt > abrt_installed_version 2>/dev/null
     rpm -q libreport > libreport_version 2>/dev/null
     rpm -q satyr > satyr_version 2>/dev/null
-    EXCLUDE_FROM_REPORT="" reporter-upload
+    echo -e "An archive will be created.\nThis archive is accessible only by ABRT developers.\nTherefore if you try to open the produced web location you will not be able to see it."
```
ernestask: >An archive will be created

IMO, that implies that the archive will be available locally, so maybe it would be better to just say something like “Uploading report data. It will not be accessible outside of the ABRT development team.”?

Also, is there actually a point in printing the URL not at debug level? Adding a secondary debug print in https://github.com/abrt/libreport/blob/779b16202c3fe990f6ecda17085c436899340e12/src/lib/curl.c#L719 for the URL makes more sense to me. And why is that a warning?

ernestask:
I’m going to merge this and then create a follow-up issue to try to get rid of some of the more useless output messages.

PR #560, opened 2019-02-26
lib: dump_dir: Clean up on failure in dd_delete()

ernestask:
The dump_dir instance is leaked if the deletion fails.

Resolves rhbz#1671233


ernestask:
Changed the code to perform the cleanup on the callee side, as I noticed that there are several codepaths where this may cause problems.

mkutlak:
LGTM :+1:

Issue #445, opened 2016-08-16
Undefined variable outside of [[ ]] bracket


kparal:
Output when reporting a new bug:

```
Bug is already reported: 1342732
Adding kparal@redhat.com to CC list
Undefined variable outside of [[ ]] bracket
```

```
<jfilak> kparal: thank you for the info ... something bad had to happen because the error message is printed out only if "reason" element is not present in the reported data
<jfilak> kparal: is this issue reproducible?
<jfilak> kparal: oh, I can reproduce it on my machine "runcon system_u:object_r:abrt_t:s0 touch /etc/foo"
<jfilak> kparal: the root cause of this issue is in the /etc/libreport/events.d/bugzilla_event.conf, the first EVENT definition lacks "-A ...."
<jfilak> kparal: I copied the "-F ..." line and replaced "-F" with "-A" and reported the AVC as https://partner-bugzilla.redhat.com/show_bug.cgi?id=1348557
<jfilak> kparal: see comment 1 and 2, I believe we don't want to include AVC description in the duplicate comments and we must replace "Description of problem:" with "Similar problem has been detected:"
<jfilak> kparal: can you please open github issue at https://github.com/abrt/libreport/issues/new
<jfilak> kparal: or file new Bugzilal bug agains libreport
<jfilak> kparal: thank you
```

PR #562, opened 2019-05-24
Add clearer warnings on missing report elements specified in fmt files

michalfabik:
This should make the immediate cause of [#445](https://github.com/abrt/libreport/issues/445) more obvious.
Of course, the reason why the files might be missing from the directories in /var/spool/abrt in the first place is a whole different problem.

### DIFF
--- a/src/lib/problem_report.c
+++ b/src/lib/problem_report.c
@@ -312,8 +312,7 @@ format_percented_string(const char *str, problem_data_t *pd, FILE *result, char 
     long len = 0;
     int opt_depth = 1;
 
-    char *missing_item = xstrdup(str);
-    memset(missing_item, 0, strlen(str));
+    char *missing_item = 0;
 
     while (*str) {
         switch (*str) {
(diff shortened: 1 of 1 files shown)


src/lib/problem_report.c
```diff
@@ -305,13 +305,16 @@ load_stream(FILE *fp)
...
 static int
-format_percented_string(const char *str, problem_data_t *pd, FILE *result)
+format_percented_string(const char *str, problem_data_t *pd, FILE *result, char *fmt_file)
 {
     long old_pos[MAX_OPT_DEPTH] = { 0 };
     int okay[MAX_OPT_DEPTH] = { 1 };
     long len = 0;
     int opt_depth = 1;
 
+    char *missing_item = xstrdup(str);
```
ernestask: I feel like this is unnecessary. What about null-initializing it here and duplicating lazily as errors are detected?

src/lib/problem_report.c
```diff
@@ -386,8 +401,10 @@ format_percented_string(const char *str, problem_data_t *pd, FILE *result)
 
     if (!okay[0])
     {
-        error_msg("Undefined variable outside of [[ ]] bracket");
+        error_msg("In format file '%s':", fmt_file);
+        error_msg("    Undefined variable '%s' outside [[ ]] brackets", missing_item);
```
ernestask: You can use `\t`.

src/lib/problem_report.c
```diff
@@ -1241,7 +1263,7 @@ problem_formatter_generate_report(const problem_formatter_t *self, problem_data_
                     self->pf_default_summary);
 
         format_percented_string(self->pf_default_summary,
-                   data, problem_report_get_buffer(pr, PR_SEC_SUMMARY));
+                   data, problem_report_get_buffer(pr, PR_SEC_SUMMARY), self->fmt_file);
```
ernestask: I had an idea for this: what if we started using `GError`s more liberally and just propagated to callers? They will surely all have a pointer to the path somewhere. The downside is that you will need to add error handling to all of them, but a cursory glance tells me that it might already be done.

ernestask: https://developer.gnome.org/glib/stable/glib-Error-Reporting.html for documentation.

ernestask: Though another concern is breaking ABI compatibility, but, in Fedora at least, no one seems to be using libreport meaningfully (aside from us).

ernestask:
Ah, didn’t notice before, but there is https://github.com/abrt/libreport/blob/ba0ee64ee10540b7cb34ed0ded7a11a60c9cc6d3/src/lib/problem_report.c#L1185, with which you can feed a format string. That means that the `fmt_file` member is null, and we end up printing a null string, which is undefined behavior, but glibc is lenient in that regard.

PR #564, opened 2019-05-28
cli: Unpack command-line argument parsing logic

ernestask:
Currently, checking for invalid command line needlessly involves
convoluted bitwise operations, which can be simplified drastically.

First, argc is reduced by optind, which points to the next argument to
be processed. If everything goes well, argc should be 1, since the only
remaining argument to be processed is the problem directory. If that
does not hold, we want to bail. Another point at which we want to bail
is when an option is passed that operates on the positional argument
(anything but -L, which just lists available events). Checking for that
involves ANDing the current option mask with the mask of all such
options. The result is NOTed for comparison, since argc is 0 in such
cases.

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -39,6 +39,9 @@ static char *steal_directory_if_needed(char *dump_dir_name)
 
 int main(int argc, char** argv)
 {
+    bool runaway_arguments;
+    bool missing_positional_argument;
+
     abrt_init(argv);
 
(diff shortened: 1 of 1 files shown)


ernestask:
This should fix a BAD_COMPARE found by covscan:

```
libreport-2.1.11/src/cli/cli.c:119: misuse_of_not: Comparing logical negation "!(opts & OPTMASK_need_arg)" against "argc" is likely to be incorrect. 
```

PR #565, opened 2019-05-29
plugins: rhbz: Don’t call strlen() on attachment data

ernestask:
It might not be textual, meaning that, if the blob somehow begins with
NUL, it would be treated as empty.

### DIFF
--- a/src/plugins/rhbz.c
+++ b/src/plugins/rhbz.c
@@ -611,7 +611,7 @@ int rhbz_attach_blob(struct abrt_xmlrpc *ax, const char *bug_id,
 {
     func_entry();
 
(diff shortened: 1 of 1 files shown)


src/plugins/rhbz.c
```diff
@@ -682,7 +682,7 @@ int rhbz_attach_fd(struct abrt_xmlrpc *ax, const char *bug_id,
 
 //TODO: need to have a method of attaching huge files (IOW: 1Gb read isn't good).
 
-    char *data = xmalloc(size + 1);
```
ernestask: I can only assume that this was done in assumption that the last byte would somehow be zero to appease `strlen()` gods.

PR #566, opened 2019-06-06
Check for empty fmt_file name when printing error msgs

michalfabik:
See commit.

### DIFF
--- a/src/lib/problem_report.c
+++ b/src/lib/problem_report.c
@@ -375,10 +375,14 @@ format_percented_string(const char *str, problem_data_t *pd, FILE *result, char 
                     fputs(item->content, result);
                     len += strlen(item->content);
                 }
-                else
+                else if (fmt_file)
+                {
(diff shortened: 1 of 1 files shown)


src/lib/problem_report.c
```diff
@@ -404,9 +408,14 @@ format_percented_string(const char *str, problem_data_t *pd, FILE *result, char
...
-                  "\tUndefined variable '%s' outside [[ ]] brackets",
-                  fmt_file, missing_item);
+        if (fmt_file)
+        {
+            error_msg("In format file '%s':\n"
+                      "\tUndefined variable '%s' outside [[ ]] brackets",
+                      fmt_file, missing_item);
+        }
+        else
+            error_msg("\tUndefined variable '%s' outside [[ ]] brackets", missing_item);
```
ernestask: I’m not sure if we want to indent here.

michalfabik: You're right, tabs removed.

PR #567, opened 2019-06-10
cli: run-command: Replace use of vfork() with fork()

ernestask:
vfork() replacing fork() was most likely an optimization, but, over
time, changes were added that violate the contract, i.e. the code
started going things other than exec and exit:
https://pubs.opengroup.org/onlinepubs/009695399/functions/vfork.html.

posix_spawn(), as recommended by Clang, would not be a suitable
replacement, as we need to call tcsetpgrp() on the child, doing which
after exec is too late if the goal is to avoid races.

### DIFF
--- a/src/cli/run-command.c
+++ b/src/cli/run-command.c
@@ -53,10 +53,10 @@ static void start_command(struct command *cmd)
 
   fflush(NULL);
 
-  cmd->pid = vfork();
+  cmd->pid = fork();
   if (cmd->pid < 0)
   {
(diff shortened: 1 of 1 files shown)


ernestask:
Another alternative could be doing the `clone()` dance manually and having a shared memory space in the end, but I think it’s a tad messier with no real benefits, as nothing else links to libreport for it to matter.

ernestask:
And, as always, this makes Clang and, in turn, Coverity happy.

PR #568, opened 2019-06-11
lib: copy_file_recursive: Use GLib abstractions

ernestask:
The current recursive copy implementation is rather cumbersome to read
and causes Coverity to complain when building RHEL packages. Using
GLib/GIO should improve readability and eliminate warnings while
retaining compatibility.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -269,7 +269,13 @@ else
 AM_CONDITIONAL(BUILD_PYTHON3, false)
 fi dnl end NO_PYTHON3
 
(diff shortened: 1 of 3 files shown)


ernestask:
To test:

1. Generate a problem directory (use your imagination)
2. `chmod -w` on the problem directory
3. `HOME=<directory that is accessible by the abrt user>` report-cli -x <problem directory>

Looking inside $HOME/.cache/abrt/spool/<dump directory>, you should see an exact copy of the original files. Current implementation chokes on the `.libreport` directory, so it’s a bit difficult to compare.

PR #570, opened 2019-07-12
dd: Update dd_get_owner to handle error return values

mkutlak:
Update dd_get_owner prototype and implementation to properly handle the
possible error return values which can be negative.

As reference the link to the issue in abrt repo:
abrt/abrt#1345

Basically dd_get_owner() calls read_number_from_file_at() which returns negative numbers in case of error:
https://github.com/abrt/libreport/blob/5e0c2c3839f7205a90aedd6be7a1f54bc4306ebb/src/lib/dump_dir.c#L240

Thank you to original commiter - webera.
https://github.com/abrt/libreport/pull/546

### DIFF
--- a/src/lib/dump_dir.c
+++ b/src/lib/dump_dir.c
@@ -737,9 +737,12 @@ int dd_set_no_owner(struct dump_dir *dd)
     return dd_set_owner(dd, no_owner_uid);
 }
 
-uid_t dd_get_owner(struct dump_dir *dd)
+uid_t dd_get_owner(struct dump_dir *dd, int *errorno)
 {
     static const long long MAX_UID_T = (1ULL << (sizeof(uid_t)*8 - 1)) - 1;
+
+    if (errorno)
+        *errorno = 0;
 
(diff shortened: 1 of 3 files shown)


mkutlak:
This will break abrt builds. WIP for now.

mkutlak:
I guess it it ready for review.

src/lib/dump_dir.c
```diff
@@ -755,8 +758,8 @@ uid_t dd_get_owner(struct dump_dir *dd)
 
     if (ret < 0)
     {
-        if (ret != -ENOENT)
-            return ret;
+        if (errorno && ret != -ENOENT)
+            *errorno = ret;
```
mgrabovsky: Thinking out loud: This is a bit different behaviour, isn't it? The previous code bailed immediately, while now only the error code is set but execution continues. It is terminated soon anyway, but the log message might be confusing.

Does it affect anything down the lane?

xsuchy:
@mgrabovsky will take over, but we will wait till @michalfabik finish his PR with locking

mgrabovsky:
Took over in #630.

PR #571, opened 2019-08-14
dirsize: Skip dirs in which sosreport is being generated

mkutlak:
With these changes ABRT will skip directories in which sosreport is running
and won't delete them when MaxCrashReportSize limit is exceeded.

I had a concern about the size of the logs that sosreport collects but
there is a limit set by sosreport for how big the individual logs can
be and the default value is 25MiB.

Fixes: rhbz#1671232


ernestask:
Is there no way to check if the file is actually currently open?

ernestask:
Nothing really standard, at least. I’ll merge and then we can deal with any fallout later (i.e. never).

PR #573, opened 2019-08-16
Wildcards in event specifications for workflows

### LINKED ISSUE #421: Custom collectors are not run in reporter-cli
Custom collectors of packages are not run after the workflows were introduced.

Also the star in `collect_*` event is not expanded in `load_event_description_from_file()`.

Star expanding was introduced in https://github.com/abrt/libreport/commit/67f19f0467f63208bab152df46bf97622e8230c0
Issue was discovered after adding https://github.com/abrt/libreport/commit/cf1bd4eb2ff931c0e9fb50d1df21099d70ca7926

mgrabovsky:
This proposal repairs the workflow (no pun intended) for workflows that contain event wildcards. The two main changes so far are:

1. Expand wildcards in the workflow XML parser so that the expanded events are available to `report-cli` and `report-gtk` alike. (I'm not happy about doing anything but parsing in the parser, but a more reasonable architecture would likely require significant refactoring.)
2. Don't fail if no matching actions are found for an event. This is a natural thing to happen when using wildcards, since some actions might only make sense for specific problem types.

Fixes #421

### DIFF
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -817,25 +817,6 @@ static int run_event_chain_real(struct run_event_state *run_state,
     return retval;
 }
 
-/**
- * Expand '*' wildcards in an event chain.
- *
- * Returns the expanded list of event names.
- */
-static GList *expand_event_chain_wildcards(GList *chain)
-{
-    GList *list = NULL;
-
-    for (GList *item = chain; item; item = g_list_next(item))
-    {
-        const char *event_name = (const char *)item->data;
-        GList *expanded = expand_event_wildcard(event_name, strlen(event_name));
-        list = g_list_concat(list, expanded);
-    }
-
-    return list;
-}
-
 /*
  * Run events from a chain. Perform the following steps for each event:
  * 1. Terminate the chain run if the backtrace is not usable.
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -813,6 +813,25 @@ static int run_event_chain_real(struct run_event_state *run_state,
     return retval;
 }
 
+/**
+ * Expand '*' wildcards in an event chain.
+ *
+ * Returns the expanded list of event names.
+ */
+static GList *expand_event_chain_wildcards(GList *chain)
+{
+    GList *list = NULL;
+
+    for (GList *item = chain; item; item = g_list_next(item))
+    {
+        const char *event_name = (const char *)item->data;
+        GList *expanded = expand_event_wildcard(event_name, strlen(event_name));
+        list = g_list_concat(list, expanded);
+    }
+
+    return list;
+}
+
 /*
  * Run events from a chain. Perform the following steps for each event:
  * 1. Terminate the chain run if the backtrace is not usable.
@@ -826,12 +845,14 @@ static int run_event_chain_real(struct run_event_state *run_state,
  */
 int run_event_chain(const char *dump_dir_name, GList *chain, int interactive)
 {
+    /* Expand *-wildcards in the names of events in the chain. */
+    GList *expanded_chain = expand_event_chain_wildcards(chain);
     struct run_event_state *run_state = new_run_event_state();
-    int retval;
 
-    retval = run_event_chain_real(run_state, dump_dir_name, chain, interactive);
+    int retval = run_event_chain_real(run_state, dump_dir_name, expanded_chain, interactive);
 
     free_run_event_state(run_state);
+    g_list_free_full(expanded_chain, free);
 
     return retval;
 }
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -796,8 +796,12 @@ static int run_event_chain_real(struct run_event_state *run_state,
             break;
         if (retval == 0 && run_state->children_count == 0)
         {
-            printf("Error: no processing is specified for event '%s'\n", event_name);
-            retval = 1;
+            /* Continue running the chain even if no actions were executed for
+             * this event.
+             */
+            log_warning("no actions matching this problem found for event '%s'",
+                    event_name);
+            continue;
         }
         else if (retval != 0 || !l_state.output_was_produced)
         {
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -782,23 +782,26 @@ static int run_event_chain_real(struct run_event_state *run_state,
     {
         l_state.output_was_produced = false;
         const char *event_name = eitem->data;
-        retval = interactive
-                ? run_event_on_dir_name_interactively(run_state, dump_dir_name, event_name)
-                : run_event_on_dir_name_batch(run_state, dump_dir_name, event_name)
-                ;
+        log_info("running commands for event '%s'", event_name);
+
+        if (interactive)
+            retval = run_event_on_dir_name_interactively(run_state, dump_dir_name,
+			    event_name);
+        else
+            retval = run_event_on_dir_name_batch(run_state, dump_dir_name,
+			    event_name);
 
         if (retval < 0)
-            /* Nothing was run (bad backtrace, user declined, etc... */
+            /* Nothing was run (bad backtrace, user declined, etc.) */
             break;
         if (retval == 0 && run_state->children_count == 0)
         {
             printf("Error: no processing is specified for event '%s'\n", event_name);
             retval = 1;
         }
-        else
-        /* If program failed, or if it finished successfully without saying anything... */
-        if (retval != 0 || !l_state.output_was_produced)
+        else if (retval != 0 || !l_state.output_was_produced)
         {
+            /* Program failed, or finished successfully with no output. */
             char *msg = exit_status_as_string(event_name, run_state->process_status);
             fputs(msg, stdout);
             free(msg);
@@ -811,15 +814,15 @@ static int run_event_chain_real(struct run_event_state *run_state,
 }
 
 /*
- * Run event from chain. Perform the following steps for each event from chain.
- * 1. Terminates a chain run if a backtrace is not usable.
- * 2. Asks for missing settings.
- * 3. Performs review of problem data if event requires review.
- * 4. Runs events commands.
- * 5. Terminates a chain run if any event from the chain requires termination
- *    (i.e. if it existed with EXIT_STOP_EVENT_RUN).
- * 6. Terminates a chain run if any error occurs.
- * 7. Continues with processing of next event.
+ * Run events from a chain. Perform the following steps for each event:
+ * 1. Terminate the chain run if the backtrace is not usable.
+ * 2. Ask for missing settings.
+ * 3. Perform review of problem data if the event requires review.
+ * 4. Run event's commands.
+ * 5. Terminate the run if any event from the chain requires termination
+ *    (i.e. if it exited with EXIT_STOP_EVENT_RUN).
+ * 6. Terminate the run if any error occurred.
+ * 7. Continue processing next event.
  */
 int run_event_chain(const char *dump_dir_name, GList *chain, int interactive)
 {
@@ -884,16 +887,17 @@ int select_and_run_workflow(const char *dump_dir_name, GHashTable *workflows, in
     {
         return -1;
     }
+
     events = wf_get_event_names(workflow);
+
     workflow_name = wf_get_name(workflow);
     environment_variable = g_strdup_printf("LIBREPORT_WORKFLOW=%s", workflow_name);
-    run_state = new_run_event_state();
 
+    run_state = new_run_event_state();
     g_ptr_array_add(run_state->extra_environment, environment_variable);
 
     retval = run_event_chain_real(run_state, dump_dir_name, events, interactive);
+    free_run_event_state(run_state);
 
     return retval;
-
-    free_run_event_state(run_state);
 }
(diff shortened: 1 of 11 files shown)


src/cli/cli.c
```diff
@@ -179,14 +179,13 @@ int main(int argc, char** argv)
...
+            /* list the workflows which can be applied to the current problem */
+            GList *possible_names = list_possible_events_glist(dump_dir_name, "workflow");
             GHashTable *possible_workflows = load_workflow_config_data_from_list(
-                                    list_possible_events_glist(dump_dir_name, "workflow"),
-                                                    WORKFLOWS_DIR);
+                            possible_names, WORKFLOWS_DIR);
+            g_list_free_full(possible_names, free);
 
-            /* this may still directory even if no workflow will be run */
+            /* this should still be a directory even if no workflow is run */
```
ernestask: s/still/steal/

src/cli/cli-report.c
```diff
@@ -782,23 +782,26 @@ static int run_event_chain_real(struct run_event_state *run_state,
...
         {
             printf("Error: no processing is specified for event '%s'\n", event_name);
             retval = 1;
         }
-        else
-        /* If program failed, or if it finished successfully without saying anything... */
-        if (retval != 0 || !l_state.output_was_produced)
+        else if (retval != 0 || !l_state.output_was_produced)
         {
+            /* Program failed, or finished successfully no output. */
```
ernestask: with no output

src/cli/cli-report.c
```diff
@@ -811,15 +814,15 @@ static int run_event_chain_real(struct run_event_state *run_state,
...
- * 1. Terminates a chain run if a backtrace is not usable.
- * 2. Asks for missing settings.
- * 3. Performs review of problem data if event requires review.
- * 4. Runs events commands.
- * 5. Terminates a chain run if any event from the chain requires termination
- *    (i.e. if it existed with EXIT_STOP_EVENT_RUN).
- * 6. Terminates a chain run if any error occurs.
- * 7. Continues with processing of next event.
+ * Run events from a chain. Perform the following steps for each event:
+ * 1. Terminate the chain run if the backtrace is not usable.
```
ernestask: Now there is a mix of definite and indefinite articles.

ernestask: (Talking about terminating runs)

src/cli/cli-report.c
```diff
@@ -828,6 +831,7 @@ int run_event_chain(const char *dump_dir_name, GList *chain, int interactive)
 
     retval = run_event_chain_real(run_state, dump_dir_name, chain, interactive);
 
+    /* Release used resources. */
```
ernestask: This is kind of obvious, no?

src/cli/cli-report.c
```diff
@@ -877,23 +881,23 @@ int select_and_run_workflow(const char *dump_dir_name, GHashTable *workflows, in
...
+
     workflow_name = wf_get_name(workflow);
     environment_variable = g_strdup_printf("LIBREPORT_WORKFLOW=%s", workflow_name);
-    run_state = new_run_event_state();
 
+    run_state = new_run_event_state();
     g_ptr_array_add(run_state->extra_environment, environment_variable);
 
-    retval = run_event_chain_real(run_state, dump_dir_name, events, interactive);
+    int retval = run_event_chain_real(run_state, dump_dir_name, events, interactive);
```
ernestask: Can we keep the declaration where it was?

src/lib/event_config.c
```diff
@@ -556,3 +556,43 @@ bool check_problem_rating_usability(const event_config_t *cfg,
...
+            while (event_files)
+            {
+                file_obj_t *file = (file_obj_t *)event_files->data;
+                const gchar *file_name = fo_get_filename(file);
+
+                /* Check if the event's name matches the required prefix. */
+                if (strncmp(file_name, event_name, event_len - 1) == 0)
+                {
+                    log_debug("found matching event '%s'", file_name);
+                    list = g_list_append(list, xstrdup(file_name));
```
ernestask: Appending is very inefficient, go for prepending and reversing at the end.

src/lib/event_config.c
```diff
@@ -556,3 +556,43 @@ bool check_problem_rating_usability(const event_config_t *cfg,
...
+                }
+
+                free_file_obj(file);
+                event_files = g_list_delete_link(event_files, event_files);
+            }
+        }
+    }
+    else
+        list = g_list_append(list, xstrdup(event_name));
```
ernestask: Might be better to invert the condition at the beginning and do something like `return g_list_prepend (NULL, xstrdup (event_name));`

src/lib/event_config.c
```diff
@@ -556,3 +556,43 @@ bool check_problem_rating_usability(const event_config_t *cfg,
 
     return result;
 }
+
+GList *expand_event_wildcard(const gchar *event_name, gsize event_len)
```
ernestask: Do we need to pass the length? What’s wrong with good old `strlen()`? Is there ever going to be a non-terminated string passed?

src/lib/event_config.c
```diff
@@ -556,3 +556,43 @@ bool check_problem_rating_usability(const event_config_t *cfg,
...
+        {
+            /* For each listed event, compare its prefix with the wildcard
+             * prefix and add it to the resulting list if it matches. */
+            while (event_files)
+            {
+                file_obj_t *file = (file_obj_t *)event_files->data;
+                const gchar *file_name = fo_get_filename(file);
+
+                /* Check if the event's name matches the required prefix. */
+                if (strncmp(file_name, event_name, event_len - 1) == 0)
```
ernestask: There is also `g_str_has_prefix()` that might be useful.

src/lib/workflow_xml_parser.c
```diff
@@ -83,16 +83,29 @@ static void text(GMarkupParseContext *context,
...
+        GList *expanded_events = expand_event_wildcard(text, text_len);
 
-        load_event_description_from_file(ec, subevent_filename);
-        if (ec_get_screen_name(ec))
-            wf_add_event(workflow, ec);
-        else
-            free_event_config(ec);
+        while (expanded_events)
+        {
+            const gchar *event_name = (const gchar *)expanded_events->data;
```
ernestask: We don’t need that explicit cast, since data pointers in `GList`s are void.

src/lib/workflow_xml_parser.c
```diff
@@ -83,16 +83,29 @@ static void text(GMarkupParseContext *context,
...
+            load_event_description_from_file(ec, event_file);
+            if (ec_get_screen_name(ec))
+            {
+                log_debug("adding event '%s' to workflow", event_name);
+                wf_add_event(workflow, ec);
+            }
+            else
+                free_event_config(ec);
+
+            free(event_file);
```
ernestask: To be honest, I would suggest adopting the automatic cleanup magic from GLib. I.e. `g_autofree char *event_file = whatever;`

ernestask: Just make sure you don’t leave them uninitialized.

src/cli/cli-report.c
```diff
@@ -826,13 +845,16 @@ static int run_event_chain_real(struct run_event_state *run_state,
  */
 int run_event_chain(const char *dump_dir_name, GList *chain, int interactive)
 {
+    /* Expand wildcards in the chain. */
```
ernestask: Comment pretty much restates  what the next line says.

src/cli/cli-report.c
```diff
@@ -826,13 +845,16 @@ static int run_event_chain_real(struct run_event_state *run_state,
...
+    GList *expanded_chain = expand_event_chain_wildcards(chain);
     struct run_event_state *run_state = new_run_event_state();
-    int retval;
 
-    retval = run_event_chain_real(run_state, dump_dir_name, chain, interactive);
+    /* Run the expanded chain. */
```
ernestask: Same here, comment adds no value.

src/cli/cli-report.c
```diff
@@ -796,8 +796,10 @@ static int run_event_chain_real(struct run_event_state *run_state,
             break;
         if (retval == 0 && run_state->children_count == 0)
         {
-            printf("Error: no processing is specified for event '%s'\n", event_name);
-            retval = 1;
+            /* Continue running the chain even if no actions were executed for
+             * this event. */
+            log_warning("no commands were run for event '%s'", event_name);
```
ernestask: No commands run or none specified?

src/lib/event_xml_parser.c
```diff
@@ -546,11 +546,15 @@ static void error(GMarkupParseContext *context,
  * event_config_t contains list of options, which is malloced by hits function
  * and must be freed by the caller
  */
-
 void load_event_description_from_file(event_config_t *event_config, const char* filename)
 {
     log_info("loading event: '%s'", filename);
-    struct my_parse_data parse_data = { {event_config, false, false, false}, {NULL, false, false}, NULL, NULL };
+    struct my_parse_data parse_data = {
```
ernestask: `my_parse_data` is just horrible, but not for this PR to fix.

src/gui-wizard-gtk/wizard.c
```diff
@@ -2155,13 +2155,13 @@ static void start_event_run(const char *event_name)
 
     if (prepare_commands(state, g_dump_dir_name, event_name) == 0)
     {
- no_cmds:
-        /* No commands needed?! (This is untypical) */
-        free_run_event_state(state);
-        char *msg = xasprintf(_("No processing for event '%s' is defined"), event_name);
+        /* No commands found for this event. Acknowledge this fact and continue
+         * with processing. */
```
ernestask: Nitpick: can you end multi-line comments with `*/` on a separate line? Makes it easier to add things.

src/gui-wizard-gtk/wizard.c
```diff
@@ -2155,13 +2155,13 @@ static void start_event_run(const char *event_name)
...
+        /* TODO: Perhaps don't even show this to the user, only log via VERB1/2 log. */
```
ernestask: I don’t know, “reacting to this situation in the GUI” will be pretty much the same, no? Although maybe without the event name.

src/gui-wizard-gtk/wizard.c
```diff
@@ -2155,13 +2155,13 @@ static void start_event_run(const char *event_name)
...
+        char *msg = xasprintf(_("No commands defined for event '%s'; skipping."), event_name);
         append_to_textview(g_tv_event_log, msg);
         free(msg);
-        cancel_processing(g_lbl_event_log, _("Processing failed."), TERMINATE_NOFLAGS);
+        /* TODO: React to this situation in the GUI. */
```
ernestask: How does it look right now without anything here? Does it just succeed?

src/gui-wizard-gtk/wizard.c
```diff
@@ -2183,7 +2183,8 @@ static void start_event_run(const char *event_name)
     if (spawn_next_command(state, g_dump_dir_name, event_name, EXECFLG_SETPGID) < 0)
     {
         unexport_event_config(env_list);
-        goto no_cmds;
+        // goto no_cmds;
```
ernestask: Why comment it out?

src/lib/workflow_xml_parser.c
mgrabovsky: OK, I'll look into that.

src/cli/cli-report.c
mgrabovsky: I felt so lonely and comment-depraved that I had to make up at least these feeble grey companions of mine here and there.

src/cli/cli-report.c
mgrabovsky: None were run because no actions matching the problem under consideration were found. I'll reword the message to make that clearer.

src/gui-wizard-gtk/wizard.c
mgrabovsky: Sure, that sounds reasonable.

src/gui-wizard-gtk/wizard.c
mgrabovsky: "WIP"

src/cli/cli-report.c
ernestask: I believe you misunderstood me - I’m talking about 1 and 6.

mgrabovsky: Indeed, you are right. Will fix.

mgrabovsky:
OK, workflows now work as expected in `report-gtk`, although I'm not completely sure about the user experience. If no actions are run for an event, the GUI lets the user know "Skipping _event_xyz_" and requires them to click "Next", possibly multiple times in a row, to proceed.

It would seem more user-friendly to simply ignore those events (log them only), skip them silently with no indication in the GUI and no interruption to the workflow.

Here is an illustration of the current state.
figure-1-report-gtk

mgrabovsky:
Ready for review now I suppose.

src/lib/event_config.c
ernestask: But that doesn’t really change anything, does it?

src/lib/event_config.c
ernestask: Which “GLib XML parser” do you have in mind?

ernestask: Aha, okay, so it’s for the actual parsing, which makes sense.

src/lib/event_config.c
ernestask: It does if the strings aren’t terminated.

PR #574, opened 2019-09-02
lib: Don't warn if a configuration key is missing

mgrabovsky:
Drop warning if a key cannot be found in the supplied configuration map
in the `try_get_map_string_item_as_*` family of functions.

Related: [abrt#1405](https://github.com/abrt/abrt/pull/1405#discussion_r318605702)

### DIFF
--- a/src/lib/abrt_types.c
+++ b/src/lib/abrt_types.c
@@ -73,7 +73,7 @@ void set_map_string_item_from_bool(map_string_t *ms, const char *key, int value)
     const char *const val_name = get_map_string_item_or_NULL(conf, item_name); \
     if (val_name == NULL) \
     { \
-        log_warning("Option '%s' is not configured", item_name); \
+        log_debug("Configuration option '%s' not found in loaded settings", item_name); \
         return 0; \
     }
 


src/lib/abrt_types.c
```diff
@@ -82,7 +81,7 @@ int try_get_map_string_item_as_bool(map_string_t *ms, const char *key, int *valu
     GET_ITEM_OR_RETURN(option, ms, key);
 
     *value = string_to_bool(option);
-    return true;
+    return 1;
```
ernestask: Unrelated change. Is this causing warnings?

ernestask:
Why is this change being made?

ernestask:
Just looking at Martin’s comment, I can’t quite tell if he meant to just remove the warning or something else entirely.

src/lib/abrt_types.c
mgrabovsky: It was just the inconsistency bugging me. There are 0s and 1s everywhere and bam, a `true` here.

mgrabovsky:
In my opinion, the warning doesn't isn't appropriate here anymore. It made sense before, when all the defaults were pulled from files. After the changes in abrt/abrt#1405, all the default values should be hardcoded and all the config files are empty by default so it's kind of expected that the function will "fail".

Martin suggested turning it into a debug-level log and perhaps reformulate the message.

ernestask:
Yeah, okay, that makes sense.

Can you somehow digest the explanation and put it in the commit message? Saves reading discussions in PRs.

src/lib/abrt_types.c
ernestask: Aha.

I would at least make this a separate commit. What do you think about changing it all to C99 booleans? Not as much mental overhead in trying to enumerate the possible return values with those.

mgrabovsky: OK, I'll perhaps leave that to later.

PR #575, opened 2019-09-18
dbus: Remove interface introspection files

mgrabovsky:
The D-Bus configuration daemon was recently removed from ABRT hence these files are not needed anymore.

Related: abrt/abrt#1405


ernestask:
Makes sense.

PR #578, opened 2019-10-15
gtk: Fix a double-free condition

mgrabovsky:
This misbehaviour seems to have been introduced in 861e8d10b.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1761095


ernestask:
Ah, I see it in the trace now.

ernestask:
The wizard code is a prime example of memory management idiocy.

/me looks into a minor refactoring.

PR #579, opened 2019-10-18
workflows: Correct name of post_report event

mgrabovsky:


ernestask:
A bit of an oops there.

PR #580, opened 2019-10-18
gtk: Memory management redux

mgrabovsky:
In light of recently (re)discovered ghastly double-frees, and for very practical reasons as well, I've taken up the task of cleaning up the wizard GUI code. I've started by flushing most of the manual memory management, which seems to be a long-time source of painful problems.

This is but a humble beginning attempting to patch the most glaring holes. There's still a lot left to do to make this into a shiny cathedral.

### DIFF
--- a/src/gui-wizard-gtk/wizard.c
+++ b/src/gui-wizard-gtk/wizard.c
@@ -377,19 +377,19 @@ static void update_window_title(void)
     /* prgname can be null according to gtk documentation */
     const char *prgname = g_get_prgname();
     const char *reason = problem_data_get_content_or_NULL(g_cd, FILENAME_REASON);
-    char *title = xasprintf("%s - %s", (reason ? reason : g_dump_dir_name),
+    g_autofree char *title = xasprintf("%s - %s", (reason ? reason : g_dump_dir_name),
             (prgname ? prgname : "report"));
     gtk_window_set_title(g_wnd_assistant, title);
-    free(title);
 }
 
 static bool ask_continue_before_steal(const char *base_dir, const char *dump_dir)
 {
-    char *msg = xasprintf(_("Need writable directory, but '%s' is not writable."
-                            " Move it to '%s' and operate on the moved data?"),
-                            dump_dir, base_dir);
+    g_autofree char *msg = xasprintf(
+            _("Need writable directory, but '%s' is not writable."
+              " Move it to '%s' and operate on the moved data?"),
+            dump_dir, base_dir);
     const bool response = run_ask_yes_no_yesforever_dialog("ask_steal_dir", msg, GTK_WINDOW(g_wnd_assistant));
-    free(msg);
+
     return response;
 }
 
@@ -400,10 +400,9 @@ struct dump_dir *wizard_open_directory_for_writing(const char *dump_dir_name)
 
     if (dd && strcmp(g_dump_dir_name, dd->dd_dirname) != 0)
     {
-        char *old_name = g_dump_dir_name;
+        free(g_dump_dir_name);
         g_dump_dir_name = xstrdup(dd->dd_dirname);
         update_window_title();
-        free(old_name);
     }
 
     return dd;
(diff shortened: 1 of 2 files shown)


src/gui-wizard-gtk/main.c
```diff
@@ -200,7 +200,7 @@ int main(int argc, char **argv)
     g_list_free_full(possible_names, free);
 
     /* if we have only 1 workflow, we can use the events from it as default */
-    if (!expert_mode && g_auto_event_list == NULL && g_hash_table_size(possible_workflows) == 1)
+    if (!expert_mode && !g_auto_event_list && g_hash_table_size(possible_workflows) == 1)
```
ernestask: Can we leave it like it was? Explicitly comparing against `NULL` at least lets you know it’s a pointer without looking at the declaration.

src/gui-wizard-gtk/wizard.c
```diff
@@ -765,12 +763,12 @@ static void tv_details_row_activated(
...
                         gpointer           user_data)
 {
-    gchar *item_name = NULL;
+    g_autofree gchar *item_name = NULL;
     struct problem_item *item = get_current_problem_item_or_NULL(tree_view, &item_name);
     if (!item || !(item->flags & CD_FLAG_TXT))
-        goto ret;
+        return;
     if (!strchr(item->content, '\n')) /* one line? */
-        goto ret; /* yes */
```
ernestask: /* no */

src/gui-wizard-gtk/wizard.c
```diff
@@ -2004,13 +1980,11 @@ static gboolean consume_cmd_output(GIOChannel *source, GIOCondition condition, g
             dd = dd_opendir(g_dump_dir_name, DD_OPEN_READONLY | DD_FAIL_QUIETLY_EACCES);
         if (!dd)
             xfunc_die();
-        char *not_reportable = dd_load_text_ext(dd, FILENAME_NOT_REPORTABLE, 0
-                                            | DD_LOAD_TEXT_RETURN_NULL_ON_FAILURE
-                                            | DD_FAIL_QUIETLY_ENOENT
-                                            | DD_FAIL_QUIETLY_EACCES);
+        g_autofree char *not_reportable = dd_load_text_ext(dd, FILENAME_NOT_REPORTABLE,
+                DD_LOAD_TEXT_RETURN_NULL_ON_FAILURE | DD_FAIL_QUIETLY_ENOENT
+                | DD_FAIL_QUIETLY_EACCES);
```
ernestask: Ouch, this really is less readable. Can you parenthesize the ORing or add newlines as previously?

src/gui-wizard-gtk/wizard.c
```diff
@@ -2116,7 +2090,7 @@ static void start_event_run(const char *event_name)
 
     if (prepare_commands(state) == 0)
     {
-        /* Strangely, no commands at all were fond for processing the crashdump.
+        /* Strangely, no commands at all were found for processing the crashdump.
```
ernestask: Not convinced that the commands are actually fond of the processing.

src/gui-wizard-gtk/wizard.c
```diff
@@ -3059,13 +3027,14 @@ static gint select_next_page_no(gint current_page_no, gpointer data)
...
-            /* Notify a user that some configuration options miss values, but */
-            /* don't force him to provide them. */
+            /* Notify a user that some configuration options miss values, but don't
+             * force him to provide them.
+             */
             check_event_config(g_event_selected);
 
-            /* >>> and this but this is clearer
-             * because it does exactly the same thing
-             * but I'm pretty scared to touch it */
```
ernestask: Aren’t we all.

ernestask:
Looks fine overall, but I would definitely put style (and comment) changes separately.

src/gui-wizard-gtk/main.c
mgrabovsky: Will do.

src/gui-wizard-gtk/wizard.c
mgrabovsky: Will try.

mgrabovsky:
I tried to keep the style changes in a separate commit, but some might have overflown.

PR #582, opened 2019-10-25
Update pull-translations script

mkutlak:
`zanata-cli` (from `zanata-platform`) isn't available for Fedora 31+.

We can use `zanata` client that comes from `python3-zanata-client` package.

### DIFF
--- a/maint/pull-translations
+++ b/maint/pull-translations
@@ -12,7 +12,14 @@ if [ -n "$(git diff-index --name-only HEAD --)" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


mgrabovsky:
Can we document somewhere that the `python3-zanata-client` package is required for this?

ernestask:
If you feel like that’s needed, go ahead.

ernestask:
Maybe checking in the script if the binary exists would be better.

mkutlak:
> Maybe checking in the script if the binary exists would be better.

Done.

PR #583, opened 2019-11-05
gui-wizard-gtk: Get rid of advanced misfeatures

ernestask:
The expert mode is completely broken and no one seems to be complaining - into the trash it goes.

The ability to user your own UI definitions is completely useless, since:
  1. we would normally have to implement changes in a backwards-compatible way,
  2. it is ridiculously easy to make everything crash and have the user not realize why,
  3. there is no valid use case, just implement a new client if you’re unhappy.

### DIFF
--- a/src/gui-wizard-gtk/main.c
+++ b/src/gui-wizard-gtk/main.c
@@ -23,7 +23,6 @@
 # include <locale.h>
 #endif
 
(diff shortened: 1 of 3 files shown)


ernestask:
@trapas, retest this please

mgrabovsky:
Seems _very_ reasonable.

PR #585, opened 2019-11-12
Fix infinite loop crash in report-gtk

mgrabovsky:


ernestask:
Can you expand on the “certain situations”?

src/gui-wizard-gtk/wizard.c
```diff
@@ -1740,7 +1740,7 @@ static gboolean consume_cmd_output(GIOChannel *source, GIOCondition condition, g
         if (is_processing_finished())
             hide_next_step_button();
         else if (retval == 0 && !g_verbose)
-            on_next_btn_cb(GTK_WIDGET(g_btn_next), NULL);
```
ernestask: This probably belongs in one of the other commits.

src/gui-wizard-gtk/wizard.c
```diff
@@ -2495,11 +2494,34 @@ static void on_page_prepare(GtkNotebook *assistant, GtkWidget *page, gpointer us
 
     if (pages[PAGENO_EVENT_PROGRESS].page_widget == page)
     {
-        if (g_event_selected && g_event_selected[0])
+        while (g_event_selected && g_event_selected[0])
         {
-            log_notice("Selected event '%s'; going to run it now...", g_event_selected);
+            log_notice("Selected event '%s'; going to run its actions now...",
+                    g_event_selected);
```
ernestask: I would keep the arguments horizontally aligned.

ernestask: Vertically? I am directionally impaired.

src/gui-wizard-gtk/wizard.c
```diff
@@ -2494,9 +2495,9 @@ static void on_page_prepare(GtkNotebook *assistant, GtkWidget *page, gpointer us
 
     if (pages[PAGENO_EVENT_PROGRESS].page_widget == page)
     {
-        log_info("g_event_selected:'%s'", g_event_selected);
         if (g_event_selected && g_event_selected[0])
         {
+            log_notice("Selected event '%s'; going to run it now...", g_event_selected);
```
ernestask: I don’t know, it feels to me like these messages should rather be logged under the debug level.

mgrabovsky: Yeah, the logging levels are all over the place. I'd place most of the new messages on info at least, although we don't really differentiate between info and debug.

I was mostly using notice here, because I wanted to see the messages, but no bother with all the other, lower-level noise.

ernestask: Oh, nevermind, I didn’t notice that `LOG_NOTICE` requires `-v`. Yeah, then probably that’s fine.

ernestask: RESOLVED WHOCARES

mgrabovsky:
Updated the last commit with gory details.

Ready for review/merge now, I suppose; although I'd love to devise a thorough refactoring of the whole thing.

ernestask:
Okay, I think this makes sense. Hopefully I’m not missing anything and this will hold up.

> although I'd love to devise a thorough refactoring of the whole thing.

If by that you mean burning it with fire, absolutely.

PR #594, opened 2020-01-28
ureport: Allow printf-like data attaching

ernestask:
Currently, it is only possible to attach string and integer data to reports (without formatting a string beforehand), with the latter case simply being a wrapper for the former.
    
This commit replaces the integer wrapper with a printf-like variadic one should there ever be a need to attach more specific kinds of data.


ernestask:
Obligatory “ABI break, yadda yadda”.

ernestask:
Ergh, realized that a variadic function would not be bindable, so the point is kind of moot.

PR #595, opened 2020-01-30
WIP: ureport: Generate GIR metadata

ernestask:
Misleadingly titled, this is a rather aggressive refactoring of the
code. In the end, it does achieve what the title suggests.

With this, we can rely on GObject Introspection and existing bindings
(PyGObject, for instance) instead of writing our own Python bindings by
hand, which is rather impossible to do without losing sanity when using
external libraries.

The refactoring bit simply turns the needed structs into boxed types,
hides the definitions behind public functions, so that we don’t poke at
the state and shoot ourselves in the foot^H^H^Hace. The tests were also
rewritten in plain C for debuggability.


ernestask:
Builds atop https://github.com/abrt/libreport/pull/594, hence the extra commit.

ernestask:
This still needs annotations on types, functions and whatever else to work properly, but it seems to be functional enough.

No, I cannot really split this up.

PR #599, opened 2020-02-13
Strata → Hydra

ernestask:
See commits.

### DIFF
--- a/src/plugins/abrt_rh_support.c
+++ b/src/plugins/abrt_rh_support.c
@@ -712,36 +712,3 @@ attach_file_to_case(const char* base_url,
     free(url);
     return result;
 }
-
-//
-// Get hint
-//
-rhts_result_t*
-get_rhts_hints(const char* base_url,
-                const char* username,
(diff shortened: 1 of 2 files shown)


mkutlak:
This should be backported to RHEL, right?

ernestask:
Ah, now I remember why I had it as a draft at first. It should be backported, yes, but Hydra isn’t live yet, so we can wait around a bit.

ernestask:
But there are some staging deployments for testing.

mkutlak:
What about the `...rs/telemetry/abrt`?

https://github.com/abrt/libreport/blob/fa0c48f32c561929691cb3b77254f3ec5b9a58f4/src/plugins/reporter-rhtsupport.c#L381

ernestask:
Oh, yes, that, too. I already complained to the people doing the migration. The idea was that they would drop the endpoint and wait for someone to complain. I complained beforehand, they wanted some information about what it’s for. Not quite sure what the outcome will be.

PR #604, opened 2020-02-26
reportclient: debuginfo: Don’t unpack RPMs directly

ernestask:
To allow dropping the setuid bit from the helper binary, this commit
reshuffles some code to extract debuginfo RPMs in a temporary directory,
add g+w permissions and move everything to the cache directory.

https://bugzilla.redhat.com/show_bug.cgi?id=1805728

### DIFF
--- a/src/client-python/reportclient/debuginfo.py
+++ b/src/client-python/reportclient/debuginfo.py
@@ -21,6 +21,7 @@
     debuginfos.
 """
 
+import stat
 import sys
 import os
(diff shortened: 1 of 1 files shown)


ernestask:
One issue with this is that `abrt-action-trim-files` would stop working, since the directories created by `cpio` would have 755 permissions. We would need to munge the spec file to chmod things, but it gets hairy with non-default locations.

PR #607, opened 2020-03-20
gui-wizard-gtk: Wrap event log messages

ernestask:
Currently, large logs received by abrt-retrace-client may cause
X/Wayland protocol errors due to an overly large buffer being allocated
for GtkTextView. Enabling word wrapping and eliminating horizontal
scrolling seems to prevent that from happening.

Cf. https://gitlab.gnome.org/GNOME/gtk/issues/1143

https://bugzilla.redhat.com/show_bug.cgi?id=1815544


mgrabovsky:
Looks lovely.

PR #608, opened 2020-03-23
rhbz: Add subcomponents info to XMLRPC params

michalfabik:
This PR adds a `sub_components` struct, as described at https://bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/WebService/Bug.html#id43 under `sub_components`, to `params` struct when reporting a bug with `reporter-bugzilla`. A list of subcomponents is retrieved from Bugzilla with a [Product.get](https://bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/WebService/Product.html) call and then checked against a hardcoded table of suitable defaults.
  
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1784490


src/plugins/bugzilla_format.conf
```diff
@@ -55,5 +55,5 @@ Additional info:: \
 
 Truncated backtrace:: %bare_%short_backtrace
 
-%attach:: -comment,-reason,-reported_to,-event_log,-extra-cc,%multiline,\
+%attach:: -comment,-sub_components,-reason,-reported_to,-event_log,-extra-cc,%multiline,\
```
ernestask: So this will attach the subcomponents that you get from Bugzilla to the report? It’s not really useful, is it? Moreover, it’s only valid for the Red Hat instance, where the extension exists.

src/lib/problem_data.c
```diff
@@ -263,6 +263,35 @@ char *problem_data_get_content_or_NULL(problem_data_t *problem_data, const char
     return item->content;
 }
 
+char *problem_data_get_sub_components(const char *component)
+{
+    const char *array[COMPONENTS_WITH_SUBCOMPONENTS][2] =
```
ernestask, line 268: You should be able to leave out the size of the first dimension unspecified. You can then use the `G_N_ELEMENTS` macro, compute it yourself or use some sort of null terminator.

ernestask:
So, as before, this only seems to be passing the value from the subcomponent file in the problem directory, which is not a good solution. Even though the name is ambiguous, “components” are primarily “packages” in the system, which are more or less fixed, while subcomponents in Bugzilla can change at any given time, so the code needs to retrieve them during the reporting.

ernestask:
https://partner-bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/Extension/SubComponents/WebService.html#bugzilla-extension-subcomponents-webservice

ernestask:
https://bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/Extension/SubComponents/WebService.html#get-subcomponents

ernestask:
Since subcomponents are mandatory where they are used, this code would fail for components that are not handled, hence the need to query Bugzilla and pick *something*.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,161 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
 
     return bugs;
 }
+
+xmlrpc_value *rhbz_get_sub_components(struct abrt_xmlrpc *ax, const char *product, const char *component)
+{
+    xmlrpc_env env;
+    xmlrpc_env_init(&env);
+
+    /* Product info is stored in a struct as follows:
```
ernestask, line 899: Can you change the formatting to this:

```
/* foo
 * bar
 * baz
 */
```
?

src/plugins/rhbz.c
```diff
@@ -874,3 +890,161 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+    {"products":({"id":i,"name":s,"description":s,"is_active":b,"default_milestone":s,"default_release":s,"has_unconfirmed":b,"classification":s,"components":({"id":i,"name":s,"description":s,"default_assigned_to":s,"default_qa_contact":s,"default_docs_contact":s,"default_cc":(s),"sort_key":i,"is_active":b,"sub_components":({})})})}
```
ernestask, line 900: Is this really needed? You could look at https://bugzilla.readthedocs.io/en/latest/api/core/v1/product.html#get-product or https://bugzilla.redhat.com/docs/en/html/api/core/v1/product.html#get-product to get the general (and current) idea of what the response would be.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,161 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+
+    where:
```
ernestask, line 902: This is covered in the link below.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,161 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+    For some reason, the same doesn't work for "components.sub_components". */
+
+    xmlrpc_value *products_array = abrt_xmlrpc_call(ax, "Product.get", "{s:(s),s:(ssss)}",
+                                                    "names", product,
+                                                    "include_fields",
+                                                    "name",
+                                                    "components.name",
+                                                    "components.sub_components",
+                                                    "components.sub_components.name");
+    if (env.fault_occurred)
```
ernestask, line 930: How does this interact with Bugzilla instances without the extension?

src/plugins/rhbz.c
```diff
@@ -874,3 +890,164 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+     * {"products":({"id":i,"name":s,"description":s,"is_active":b,"default_milestone":s,"default_release":s,"has_unconfirmed":b,"classification":s,"components":({"id":i,"name":s,"description":s,"default_assigned_to":s,"default_qa_contact":s,"default_docs_contact":s,"default_cc":(s),"sort_key":i,"is_active":b,"sub_components":({})})})}
+     *
+     * where:
+     * () = array
+     * {} = struct
+     * b  = bool
+     * i  = int
+     * s  = string
+     * 
+     * The structs in the "sub_components" array are almost identical to the ones in the parent
```
ernestask, line 909: Once the code for Red Hat Bugzilla is released to the public, this should just point there.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,164 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+     * subcomponents.
+     * 
+     * http://xmlrpc-c.sourceforge.net/doc/libxmlrpc.html#formatstring
+     * 
+     * First, obtain a struct containing one item called "products" whose value is an array containing one struct
+     * containing information about the product we specified in the function arg. The "include_fields" arg
+     * specifies the struct as the following subset of the above:
+     * {"products":({"name":s,"components":({"name":s,"sub_components":({})})})}
+     * 
+     * The "components" struct doesn't have to be specified by itself because we specify its members.
```
ernestask, line 920: I’m really struggling to understand this. Specified where? You’re looking at products here and accessing their components.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,164 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+     * For some reason, the same doesn't work for "components.sub_components".
+     */
+
+    xmlrpc_value *products_array = abrt_xmlrpc_call(ax, "Product.get", "{s:(s),s:(ssss)}",
+                                                    "names", product,
+                                                    "include_fields",
+                                                    "name",
```
ernestask, line 927: Do you need the product name?

src/plugins/rhbz.c
```diff
@@ -874,3 +890,164 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+    xmlrpc_value *comp_array = rhbz_get_member("components", product_struct);
+    if (env.fault_occurred)
+        abrt_xmlrpc_die(&env);
+
+    unsigned array_size = rhbz_array_size(comp_array);
+
+    const char *str_comp_name;
+    xmlrpc_value *comp_struct = NULL;
+
+    xmlrpc_value *comp_nameV = NULL;
```
ernestask, line 956: I think it would make more sense to limit the scope and have the cleanup be as close to allocation as possible. You don’t use this outside of the loop.

ernestask: Bonus points for not using the Hungarian notation.

src/plugins/rhbz.c
```diff
@@ -874,3 +890,164 @@ xmlrpc_value *rhbz_search_duphash(struct abrt_xmlrpc *ax,
...
+    if (env.fault_occurred)
+        abrt_xmlrpc_die(&env);
+
+    xmlrpc_value *comp_array = rhbz_get_member("components", product_struct);
+    if (env.fault_occurred)
+        abrt_xmlrpc_die(&env);
+
+    unsigned array_size = rhbz_array_size(comp_array);
+
+    const char *str_comp_name;
```
ernestask, line 953: The author of XMLRPC-C must have been smoking some real good stuff with the placement of const-qualifiers.

src/plugins/rhbz.c
michalfabik: Specified in the "include_fields" array that you're passing as an arg to `abrt_xmlrpc_call`.  
If you do `"{s:(s)}", "include_fields", "components.name"`, you get the product struct conaining the `components` struct, which cointains the `name`. I.e. you don't have to specify that you want the `components` struct because it's already mentioned in `components.name`. However, if you attempt the same with the `sub_components` struct (i.e. `{s:(ss)}, "include_fields", "components.name", "components.sub_components.name"`) you don't get the `sub_components` struct at all (or you get it empty, I don't remember). You have to explicitly include `"components.sub_components"` _and_ `"components.sub_components.name"` as args to the format string.  
  
Anyway, I removed most of the long comments with the struct description.

src/plugins/rhbz.c
michalfabik: Actually no, you're right.

src/plugins/rhbz.c
michalfabik: Innit! That's what struck me as odd too.

src/plugins/rhbz.c
michalfabik: Yeah I know, I'm always running out of ideas in situations like this.

src/plugins/rhbz.c
ernestask: Aha! This is interesting.

https://bugzilla.redhat.com/docs/en/html/integrating/api/Bugzilla/WebService/Product.html


    sub_components

    array of hashes. This value must be specified in the include_fields or extra_fields argument. This feature is a Red Hat customisation.

Maybe that’s a clue.

ernestask: In any case, can you replace the text with this dumbed-down version? I’ll be scratching my head sometime in the future when I read this again.

ernestask:
Could you also look into writing some tests? You could write your own mock version of Bugzilla like here: https://github.com/cockpit-project/cockpit/blob/master/test/verify/files/mock-bugzilla-server.py

src/plugins/rhbz.c
michalfabik: Rewritten, hopefully for the better.

src/plugins/rhbz.c
michalfabik: There. I still don't like it, but purely as for Hungarianness it should be a little better.  
  
Cleanup moved too.

michalfabik:
> Could you also look into writing some tests? You could write your own mock version of Bugzilla like here: https://github.com/cockpit-project/cockpit/blob/master/test/verify/files/mock-bugzilla-server.py  
  
Oh, nice. I'll have a look at it.

src/plugins/rhbz.c
michalfabik: I added a check to make sure we only retrieve the subcomponents for product "Red Hat Enterprise Linux <X>". With product "Fedora", the query for subcomponent fails after a _long_ time of waiting. I wonder if it's possible to find out the info that product so-and-so requires subcomponents directly, without the `Product.get` call.

ernestask: Probably because there is an exorbitant amount of components for the Fedora product. Instead of using hacks, I would try to reach out to the maintainer of Red Hat Bugzilla. Likely this will require using an extension to query a component directly. This will need changing regardless because of that detail.

With the above, the answer to my original question would be “not very”, so we will need handling to not explode should someone use a more conventional Bugzilla instance.

michalfabik: So I [wrote to RH BZ maintainers](https://redhat.service-now.com/help?id=rh_ticket&table=sc_req_item&sys_id=a8c8c218db705c90c0c8464e1396192e), (probably using a completely wrong communication channel). Thoughts?

michalfabik:
Some tests added, some are probably unnecessary but I wasn't sure if I was even doing it right. I might remove some of it or use it as a basis for more testing of the XMLRPC communication. Please don't shoot me.

src/plugins/rhbz.c
```diff
@@ -941,3 +941,69 @@ xmlrpc_value *rhbz_get_sub_components(struct abrt_xmlrpc *ax, const char *produc
...
+            {"test",          "sub1"},
+            {"valgrind",      "system-version"},
+            {"virtio-win",    "distribution"}
+    };
+    char *candidate = NULL;
+
+    for (int i = 0; i < G_N_ELEMENTS(defaults); i++)
+        if (strcmp(component, defaults[i][0]) == 0)
+        {
+            candidate = strdup(defaults[i][1]);
```
ernestask, line 997: Why the `strdup()`?

tests/rhbz_functions.at
```diff
@@ -0,0 +1,275 @@
...
+        if (env.fault_occurred)
+            printf("Error in env_init: %s\n", env.fault_string);
+        xmlrpc_client_setup_global_const(&env);
+        if (env.fault_occurred)
+            printf("Error in setup_global_const: %s\n", env.fault_string);
+
+        struct abrt_xmlrpc *client = abrt_xmlrpc_new_client(rhbz.b_bugzilla_xmlrpc, rhbz.b_ssl_verify);
+
+        {
+            xmlrpc_value *result = abrt_xmlrpc_call_full(&env, client, "User.login", "{s:s,s:s}",
```
ernestask, line 124: I’m not sure we want to test these functions directly, because they are very thing wrappers for XMLRPC-C. Why not `rhbz_login()`?

tests/rhbz_functions.at
```diff
@@ -0,0 +1,275 @@
...
+            xmlrpc_value *sub_components = rhbz_get_sub_components(client, "Fedora", "gdb");
+            TS_ASSERT_PTR_IS_NOT_NULL(sub_components);
+            if (sub_components)
+                xmlrpc_DECREF(sub_components);
+        }
+
+        {
+            xmlrpc_value *sub_components = rhbz_get_sub_components(client, "Red Hat Enterprise Linux 8", "gdb");
+            TS_ASSERT_PTR_IS_NOT_NULL(sub_components);
+            const char *sub_component = rhbz_get_default_sub_component("gdb", sub_components);
```
ernestask, line 238: No need for the const-qualifier.

tests/rhbz_functions.at
```diff
@@ -0,0 +1,275 @@
...
+TS_MAIN
+{
+    pid_t pid = fork();
+
+    if (pid == 0)
+    {
+        int ret = 0;
+        char *argv[2];
+        argv[0] = (char *)"../../mock_bugzilla_server/mock_bugzilla_server.py";
+        argv[1] = NULL;
```
ernestask, line 57:
```suggestion
        char *argv[2] =
        {
            (char *)"../../mock_bugzilla_server/mock_bugzilla_server.py";
            NULL,
        };
```

ernestask: Though I would personally cast the array itself in the call to `execv()`.

tests/mock_bugzilla_server/mock_bugzilla_server.py
```diff
@@ -0,0 +1,208 @@
...
+            def process_include_extra_fields(field_list):
+                field_list = sorted(set(field_list))
+                for key in extra_components_fields:
+                    if 'components.%s' % key not in field_list:
+                        for f in field_list:
+                            if f.startswith('components.%s' % key):
+                                while f in field_list:
+                                    field_list.remove(f)
+
+
```
ernestask, line 124: Some stray whitespace.

tests/rhbz_functions.at
```diff
@@ -0,0 +1,275 @@
...
+
+        {
+            xmlrpc_value *sub_components = rhbz_get_sub_components(client, "Fedora", "afmw");
+            TS_ASSERT_PTR_IS_NULL(sub_components);
+            if (sub_components)
+                xmlrpc_DECREF(sub_components);
+        }
+
+        /* Should succeed even though we don't want to report sub_components for Fedora
+         * because the product name check is up in rhbz_new_bug. Do we want to move it one level down
```
ernestask, line 225: I don’t think we do. The result would be, what, an empty value? That sounds fine to me.

ernestask:
And don’t remove the POT file with this.

ernestask:
Looks fine in general. Not sure we can get much elegance with the current state of the code and lack of support in Red Hat Bugzilla, so I’m saying +1.

The tests cover a bit too much, IMO. I would try to replicate reporter-bugzilla for now instead of testing what subcomponents the mock server returns.

src/plugins/rhbz.c
michalfabik: I don't remember any more. Probably just a leftover from an earlier working version of the code. Removed.

tests/rhbz_functions.at
michalfabik: Rewritten, though casting in the call (instead of in the definition of `argv`) makes the compiler complain about a discarded `const` in the assignment.

michalfabik:
> And don’t remove the POT file with this.  
  
Restored, sorry.

tests/rhbz_functions.at
ernestask: Well, the idea is that you make the array const (and you can drop the size, really). It’s only the `execv()` that is idiosyncratic.

michalfabik: Ah, thanks. Rewritten.

michalfabik:
> I would try to replicate reporter-bugzilla for now instead of testing what subcomponents the mock server returns.  
  
I removed some of the code that dealed directly with the subcomponents from the tests and added tests for all `rhbz_` functions that are called from `reporter-bugzilla` directly and a few lower-level ones that I mostly used for evaluating the results.

PR #610, opened 2020-04-01
Support new kernel namespace

ernestask:
Linux 5.6 introduces a new time namespace:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=769071ac9f20b6a447410c7eaa55d1a5233ef40c

### DIFF
--- a/src/include/internal_libreport.h
+++ b/src/include/internal_libreport.h
@@ -729,7 +729,19 @@ int get_env_variable(pid_t pid, const char *name, char **value);
 #define PROC_NS_ID_USER 4
 #define PROC_NS_ID_UTS 5
 #define PROC_NS_ID_CGROUP 6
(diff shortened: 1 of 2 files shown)


src/include/internal_libreport.h
```diff
@@ -729,7 +729,19 @@ int get_env_variable(pid_t pid, const char *name, char **value);
...
+    "ipc",
+    "mnt",
+    "net",
+    "pid",
+    "uts",
+    "user",
+    "cgroup",
+    "pid_for_children",
+    "time",
+    "time_for_children",
```
mgrabovsky, line 743: I might be overly pedantic – would it make sense to have them sorted in alphabetic order?

ernestask: Well, my concern here is binary compatibility, what with the internal header not being internal. And it does seem that ABRT uses some of the indices defined above, so we would need to rebuild it.

ernestask: Actually, since we’re already breaking it with Michal’s changes, I might as well reorder these.

PR #620, opened 2020-04-19
Use POSIX expr in place of bashisms

EvaSDK:
expr tested with GNU coreutils and Mac OSX BSD util.
Shell substitution tested with dash.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -116,8 +116,8 @@ AC_SUBST(XMLRPC_CLIENT_LIBS)
 # enable bugzilla & deps translations
 for FILE in `grep -e "#.*ugzilla.*" -e "#.*naconda.*" po/POTFILES.in`
 do
-  sed -ie "s,$FILE,${FILE:1}," po/POTFILES.in
-  sed -ie "\,^${FILE:1}$,d" po/POTFILES.skip
+  sed -ie "s,$FILE,${FILE#\#}," po/POTFILES.in
+  sed -ie "\,^${FILE#\#}$,d" po/POTFILES.skip
 done
 else
 AM_CONDITIONAL(BUILD_BUGZILLA, false)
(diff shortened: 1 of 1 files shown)


ernestask:
Thanks for the contribution! Can you update the commit message to specify the reasons for this change? Also change the subject line to look like “configure.ac: Foo the bars”.

ernestask:
@abrt-bot, test this please

EvaSDK:
Hopefully this is better now.

configure.ac
```diff
@@ -116,16 +116,16 @@ AC_SUBST(XMLRPC_CLIENT_LIBS)
...
+  sed -ie "\,^${FILE#\#}$,d" po/POTFILES.skip
 done
 else
 AM_CONDITIONAL(BUILD_BUGZILLA, false)
 
 # disablie bugzilla & deps translations
 for FILE in `grep -e "ugzilla" -e "naconda" po/POTFILES.in`
 do
-  if test "${FILE:0:1}" = "#"
+  if test `expr "${FILE}" : .` = "#"
```
ernestask, line 128: Does this actually work as intended? Pattern-matching in expr seems to return the number of matched characters, so this would result in `test 1 = "#"`, which is obviously false for anything. Moreover, wouldn’t `$()` in this case be more POSIX-compliant?

ernestask:
```suggestion
  if test $(expr "$FILE" : "#") != '0'
```

ernestask: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/expr.html

EvaSDK: About expr, you are correct, my tests where done with string 123
so `expr "123" : .`, `expr "123" : ..` or `expr "123" : ...` would return something that looked correct. Running `expr "$FILE" : #` should be sufficient as it returns 1 or 0 according to whether it matched or not.

About `$()` vs `\`\``, I don't know which is more POSIX but dash supports both so I took the approach of modifying the least possible things.

EvaSDK: Updated patch without test since command exist status is what we need now.

ernestask: Oh, actually both `$()` and ``` `` ``` are fine. The more I know. Not needed with your change anyway. :p

PR #622, opened 2020-04-22
Allow editing CHANGELOG.md before release commit, drop pull-translations

michalfabik:
`$ make release-(major|minor|bug)` creates a new empty entry in CHANGELOG.md
but commits the change before the user has a chance to fill it, making
it necessary to ammend and re-tag the release commit. This commit opens
an editor to let the user edit the changelog in time.  
  
Also, remove `maint/pull-translations` which was only relevant to Zanata.


Makefile.am
```diff
@@ -68,9 +68,14 @@ release:
...
 	sed "$$(grep -n %changelog libreport.spec.in | cut -f1 -d: | head -1)"'r /tmp/changelog.tmp' -i libreport.spec.in; \
-	sed -e "s/^## \[Unreleased\]/## [Unreleased]\n\n## [$$NEW_VER] - $$(date +'%F')/" \
+	sed -e "s/^## \[Unreleased\]/## [Unreleased]\n\n## [$$NEW_VER] - $$(date +'%F')\n[\/\/] # (Please sum up added, changed, removed and fixed features here. See below for examples.)/" \
 	    -e "s/^\[Unreleased\]: \(https:\/\/.*\/compare\)\(\/.*\)\.\.\.HEAD/[Unreleased]: \1\/$$NEW_VER...HEAD\n[$$NEW_VER]: \1\2...$$NEW_VER/" \
-	    -i CHANGELOG.md; \
+	    -i CHANGELOG.md
+	if [ -z $$EDITOR ]; then \
+		/usr/bin/vi CHANGELOG.md; \
+	else \
+		$(EDITOR) CHANGELOG.md; \
```
ernestask, line 77: You can replace this with a one-liner: `${EDITOR:-vi} CHANGELOG.md`.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

ernestask: For completeness, `VISUAL` should also be consulted, but ehhhhhhhhhhhhhhhhhhhhhhhhhh.

michalfabik: > You can replace this with a one-liner  
  
Wow, nice, I had no idea this exists. Replaced.

Makefile.am
```diff
@@ -68,9 +68,10 @@ release:
 	echo "* $$(LC_TIME='en_US.UTF-8' date +'%a %b %d %Y') $$(git config --get user.name) <$$(git config --get user.email)> $$NEW_VER-1" | sort > /tmp/changelog.tmp; \
 	git log --oneline $$OLD_VER..HEAD | awk '{$$1=""; if (a[$$0]++ == 0) print "-" $$0} END {print ""}' | grep -v -e "- Merge" -e "- testsuite:" -e "- make:" >> /tmp/changelog.tmp; \
 	sed "$$(grep -n %changelog libreport.spec.in | cut -f1 -d: | head -1)"'r /tmp/changelog.tmp' -i libreport.spec.in; \
-	sed -e "s/^## \[Unreleased\]/## [Unreleased]\n\n## [$$NEW_VER] - $$(date +'%F')/" \
+	sed -e "s/^## \[Unreleased\]/## [Unreleased]\n\n## [$$NEW_VER] - $$(date +'%F')\n[\/\/] # (Please sum up added, changed, removed and fixed features here. See below for examples.)/" \
 	    -e "s/^\[Unreleased\]: \(https:\/\/.*\/compare\)\(\/.*\)\.\.\.HEAD/[Unreleased]: \1\/$$NEW_VER...HEAD\n[$$NEW_VER]: \1\2...$$NEW_VER/" \
-	    -i CHANGELOG.md; \
```
ernestask, line 73: You might need to keep the escaping (and use it in the following line), since each command is run in a separate shell, so later expansions of `NEW_VER` and friends would end up being empty.

michalfabik: That's strange, I tried running `make release-minor` repeatedly and the version e.g. in `libreport-version` and `gen-version` keeps incrementing like it should.  
I tried with the escaping at the beginning but when the editor opened, the changes weren't yet there.

ernestask: Yeah, you’re right, this really doesn’t make sense. I think it was only relevant for 27c861ad3dabae0d64f35ba965370266b807c2d4, because it was setting environment variables in the script itself, but, with the current setup, you either need to set `NEW_VER` yourself for the `release` target (and that will remain available until the process exits) or `release-{fix,minor,major}` will recursively run `make` and set variables for the invocation.

So, I think we can do it the other way around and remove the continuations?

michalfabik: Removed.

PR #623, opened 2020-04-22
lib: Remove creates-items tag parsing in event definitions

ernestask:
c9120cdc51ce7aba5a19500bf24692144ac64983 in abrt introduced these for
the reporting wizard (advanced mode), but, with that gone, this makes no
sense to keep around.


ernestask:
Marking as draft until I create a PR in abrt to clean up event definitions.

PR #626, opened 2020-04-29
Compile warnings

mzidek-gh:
This adds some missing doc strings in python bindings and fixes a bunch of warnings during compilation.

### DIFF
--- a/src/cli/run-command.c
+++ b/src/cli/run-command.c
@@ -146,7 +146,7 @@ static int finish_command(struct command *cmd)
 
 int run_command(char **argv)
 {
-  struct command cmd = { argv, 0 };
+  struct command cmd = { .argv = argv, .pid = 0 };
   start_command(&cmd);
   return finish_command(&cmd);
 }
--- a/src/client-python/reportclient/clientmodule.c
+++ b/src/client-python/reportclient/clientmodule.c
@@ -23,13 +23,13 @@
 static PyMethodDef module_methods[] = {
     /* method_name, func, flags, doc_string */
     /* for include/client.h */
-    { "alert"                     , p_alert                   , METH_VARARGS },
-    { "ask"                       , p_ask                     , METH_VARARGS },
-    { "ask_password"              , p_ask_password            , METH_VARARGS },
-    { "ask_yes_no"                , p_ask_yes_no              , METH_VARARGS },
-    { "ask_yes_no_yesforever"     , p_ask_yes_no_yesforever   , METH_VARARGS },
-    { "ask_yes_no_save_result"    , p_ask_yes_no_save_result  , METH_VARARGS },
-    { NULL }
+    { "alert"                     , p_alert                   , METH_VARARGS, NULL },
+    { "ask"                       , p_ask                     , METH_VARARGS, NULL },
+    { "ask_password"              , p_ask_password            , METH_VARARGS, NULL },
+    { "ask_yes_no"                , p_ask_yes_no              , METH_VARARGS, NULL },
+    { "ask_yes_no_yesforever"     , p_ask_yes_no_yesforever   , METH_VARARGS, NULL },
+    { "ask_yes_no_save_result"    , p_ask_yes_no_save_result  , METH_VARARGS, NULL },
+    { NULL, NULL, 0, NULL }
 };
 
 #if PY_MAJOR_VERSION >= 3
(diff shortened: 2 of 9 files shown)


ernestask:
@abrt-bot, add to whitelist

ernestask:
The commit messages need some work:

  - Don’t use the file name for code changes in the subject; the commit log should get you on the right path
  - The bodies lack details (what warnings? what was the initializer missing?)

src/report-python/report/reportmodule.c
```diff
@@ -44,14 +44,14 @@ PyObject *ReportError;
...
-    { "delete_dump_dir"           , p_delete_dump_dir         , METH_VARARGS },
+    { "dd_opendir", p_dd_opendir, METH_VARARGS, "Opens the given path"},
+    { "dd_create", p_dd_create, METH_VARARGS, "Create a dump directory"},
+    { "delete_dump_dir", p_delete_dump_dir, METH_VARARGS, "Delete a dump directory" },
     /* for include/report/report.h */
-    { "report_problem_in_dir"     , p_report_problem_in_dir   , METH_VARARGS },
-    { "report_problem_in_memory"  , p_report_problem_in_memory, METH_VARARGS },
-    { "report_problem"            , p_report_problem          , METH_VARARGS },
-    { NULL }
+    { "report_problem_in_dir", p_report_problem_in_dir, METH_VARARGS, "Reports a problem stored in a directory. Allows to specify flags."},
```
ernestask, line 51: So it allows specifying flags… Is it a positional argument? A keyword argument? Reading this would rather make me go read the code and ignore the docstrings.

mzidek-gh:
I have changed this PR to only deal with the -Wmissing-field-initializers and do not add any doctext here, so it is more focused. There are several places where the doctexts could be added so it is better to do that in a separate PR (and I can not come up with good wording for some of them right now, so I want to clean up at least the warnings and do the doctexts later :) )

So there is just one patch now.

src/client-python/reportclient/clientmodule.c
```diff
@@ -23,13 +23,13 @@
 static PyMethodDef module_methods[] = {
     /* method_name, func, flags, doc_string */
     /* for include/client.h */
-    { "alert"                     , p_alert                   , METH_VARARGS },
```
ernestask, line 26: If you’re going to change the style, do it separately.

ernestask:
Don’t initializer lists without designations result in the same thing?

ISO/IEC 9899:TC3 6.7.8.19:
The initialization shall occur in initializer list order, each initializer provided for a
particular subobject overriding any previously listed initializer for the same subobject; <sup>132)</sup>
all subobjects that are not initialized explicitly shall be initialized implicitly the same as
objects that have static storage duration.

I think the warning means something else - that the programmer forgot to explicitly initialize the fields.

src/cli/run-command.c
```diff
@@ -146,7 +146,7 @@ static int finish_command(struct command *cmd)
 
 int run_command(char **argv)
 {
-  struct command cmd = { argv, 0 };
+  struct command cmd = { .argv=argv, .pid=0 };
```
ernestask, line 149: Same thing, if you are going to add designations just for the sake of having them, do it separately.

We also surround operators with whitespace (and I would prefer seeing initializers being on their own lines).

src/client-python/reportclient/clientmodule.c
mzidek-gh: Ah, sorry, I did not want to change the coding style here. Just add the NULL at the end. I wrongly merged two WIP patches into this one (that is why also the docstrings were still present, I removed them in the latest versions)

src/cli/run-command.c
mzidek-gh: I added the whitespaces. But I do not add the designations just for the sake of having them, this is actually what I do in this patch, either add designations or init all members explicitly. If we only init several (not all) fields without designations it is going to trigger the warning that I try get rid of with this patch.

mzidek-gh:
> Don’t initializer lists without designations result in the same thing?
> 
> ISO/IEC 9899:TC3 6.7.8.19:
> The initialization shall occur in initializer list order, each initializer provided for a
> particular subobject overriding any previously listed initializer for the same subobject; 132)
> all subobjects that are not initialized explicitly shall be initialized implicitly the same as
> objects that have static storage duration.
> 
> I think the warning means something else - that the programmer forgot to explicitly initialize the fields.

Yes, that is why I mostly add the missing fields explicitly with the few exceptions where I think it makes sense to use the designations (like the struct command structure).

mzidek-gh:
All changed lines were changed for the sake of not generating the warning, not for the sake of changing the coding style (in the current version). The designations are hint for the compiler that some fields were knowingly omitted so it does not generate the warning.

src/cli/run-command.c
ernestask: Yeah, you are absolutely correct. I have no idea why I said that, sorry.

PR #627, opened 2020-04-29
Add a commit message template

mzidek-gh:
To start using this template:
$ git config commit.template .git-commit-template

If this makes sense, we could add commit templates to other abrt repositories as well.

### DIFF
--- a/.git-commit-template
+++ b/.git-commit-template
@@ -1,0 +1,13 @@
+COMPONENT: Subject
+
+Explanation
+
+Resolves:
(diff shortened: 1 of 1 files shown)


michalfabik:
Nice, LGTM.

ernestask:
This should be mentioned somewhere (probably HACKING.md).

michalfabik:
> This should be mentioned somewhere (probably HACKING.md).  
  
Done.

PR #629, opened 2020-05-04
Don't pick .lock'd and new dirs for deletion when problem dirs grow over size limit

michalfabik:
See commits.

Resolves https://github.com/abrt/abrt/issues/1450.  
  
See also https://github.com/abrt/abrt/pull/1481

### DIFF
--- a/src/lib/dirsize.c
+++ b/src/lib/dirsize.c
@@ -91,22 +91,24 @@ double libreport_get_dirsize_find_largest_dir(
     {
         if (libreport_dot_or_dotdot(ep->d_name))
             continue;
-        char *dname = libreport_concat_path_file(pPath, ep->d_name);
-        if (lstat(libreport_concat_path_file(dname, "sosreport.log"), &statbuf) == 0)
+        g_autofree char *dname = libreport_concat_path_file(pPath, ep->d_name);
+        g_autofree char *sosreport_path = libreport_concat_path_file(dname, "sosreport.log");
+        g_autofree char *lock_path = libreport_concat_path_file(dname, ".lock");
+        if (lstat(sosreport_path, &statbuf) == 0)
         {
             log_debug("Skipping %s': sosreport is being generated.", dname);
             size += libreport_get_dirsize(dname);
-            goto next;
+            continue;
         }
-        if (lstat(libreport_concat_path_file(dname, ".lock"), &statbuf) == 0)
+        if (lstat(lock_path, &statbuf) == 0)
         {
             log_warning("Skipping %s: directory locked. Is a backtrace being generated?", dname);
             size += libreport_get_dirsize(dname);
-            goto next;
+            continue;
         }
         if (lstat(dname, &statbuf) != 0)
         {
-            goto next;
+            continue;
         }
         if (S_ISDIR(statbuf.st_mode))
         {
@@ -145,8 +147,6 @@ double libreport_get_dirsize_find_largest_dir(
         {
             size += statbuf.st_size;
         }
- next:
-        free(dname);
     }
     closedir(dp);
     return size;
--- a/src/lib/dirsize.c
+++ b/src/lib/dirsize.c
@@ -72,7 +72,8 @@ static bool this_is_a_dd(const char *dirname)
 double libreport_get_dirsize_find_largest_dir(
         const char *pPath,
         char **worst_dir,
-        const char *excluded)
+        const char *excluded,
+        const char *proc_dir)
 {
     if (worst_dir)
         *worst_dir = NULL;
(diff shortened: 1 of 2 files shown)


src/lib/dirsize.c
```diff
@@ -130,9 +131,12 @@ double libreport_get_dirsize_find_largest_dir(
                     }
                     else
                     {
-                        maxsz = sz;
-                        free(*worst_dir);
-                        *worst_dir = libreport_xstrdup(ep->d_name);
+                        if (strcmp(ep->d_name, proc_dir))
```
mgrabovsky, line 134: This segfaults if `proc_dir == NULL`.

Also, I'd prefer to have the explicit `!= 0` there.

src/lib/dirsize.c
```diff
@@ -96,6 +97,13 @@ double libreport_get_dirsize_find_largest_dir(
             log_debug("Skipping %s': sosreport is being generated.", dname);
             goto next;
         }
+        if (lstat(libreport_concat_path_file(dname, ".lock"), &statbuf) == 0)
```
ernestask, line 100: Aren’t you leaking the string here?

src/lib/dirsize.c
```diff
@@ -96,6 +97,13 @@ double libreport_get_dirsize_find_largest_dir(
...
+        {
+            log_warning("Skipping %s: directory locked. Is a backtrace being generated?", dname);
+            double sz = libreport_get_dirsize(dname);
```
ernestask, line 103: I’m not sure the intermediate variable is really necessary in this case.

src/lib/dirsize.c
```diff
@@ -96,6 +97,13 @@ double libreport_get_dirsize_find_largest_dir(
...
+            size += sz;
```
ernestask, line 104: Why is this needed if we’re skipping the directory? I’m looking at the sosreport-still-being-generated case above and it doesn’t do that.

michalfabik: I thought the directory's size still counts to the total size of the directory where problem directories are stored `/var/spool/abrt`, even if we're skipping it for now.

ernestask: Yeah, that makes sense. Do you think we should do the same above, then?

ernestask: *Ideally*, we would fold the two cases into yours, but that might be more involved, given that sosreport generation comes from an event handler.

src/lib/dirsize.c
ernestask: Could we also use `g_build_filename()` here?

src/lib/dirsize.c
michalfabik: > Yeah, that makes sense. Do you think we should do the same above, then?  
  
Yeah, makes sense to me.

src/lib/dirsize.c
michalfabik: > Aren’t you leaking the string here?  
  
`dname`? That one is `free()`'d at `next:`.

michalfabik: > Could we also use `g_build_filename()` here?  
  
Oh, absolutely, but I'd rather do it in a separate pull request along with all the other GLib migration changes.

ernestask: > Oh, absolutely, but I'd rather do it in a separate pull request along with all the other GLib migration changes.

Okay, just throwing it out there to make it one line less when you do that.

ernestask: > `dname`? That one is `free()`'d at `next:`.

`libreport_concat_path_file()` returns a heap-allocated string, which should be freed.

michalfabik: Leaks fixed.

src/lib/dirsize.c
```diff
@@ -147,6 +149,8 @@ double libreport_get_dirsize_find_largest_dir(
         }
  next:
         free(dname);
+        free(sosreport_path);
+        free(lock_path);
```
mgrabovsky, line 153: I'd recommend to use `g_autofree` instead. (Thanks, Ernestas.)

michalfabik: Right, I actually thought about this right after I pushed the commit. :\

michalfabik: Like this? And we got rid of the `goto`, too.

mgrabovsky: Yup, looks better now :+1:

src/lib/dirsize.c
```diff
@@ -123,9 +133,12 @@ double libreport_get_dirsize_find_largest_dir(
                     }
                     else
                     {
-                        maxsz = sz;
-                        free(*worst_dir);
-                        *worst_dir = libreport_xstrdup(ep->d_name);
+                        if (!proc_dir || strcmp(proc_dir, ep->d_name) != 0)
```
ernestask, line 136: You could fold this into an `else if` and save some horizontal whitespace.

ernestask:
Looks good to me overall.

PR #630, opened 2020-05-06
dd: Update dd_get_owner to handle error return values

mgrabovsky:
Takeover from #570.

Pretty much just rebased on top of master.

Follow-up: abrt/abrt#1484

### DIFF
--- a/src/lib/dump_dir.c
+++ b/src/lib/dump_dir.c
@@ -744,6 +744,7 @@ uid_t dd_get_owner(struct dump_dir *dd)
     int dd_md_fd = dd_get_meta_data_dir_fd(dd, /*no create*/0);
     if (dd_md_fd < 0)
     {
+        errno = 0;
         log_info("No meta-data, using fs owner.");
         return dd->dd_uid;
     }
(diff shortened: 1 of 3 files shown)


src/lib/dump_dir.c
```diff
@@ -755,10 +758,11 @@ uid_t dd_get_owner(struct dump_dir *dd)
 
     if (ret < 0)
     {
-        if (ret != -ENOENT)
-            return ret;
+        if (errorno && ret != -ENOENT)
+            *errorno = ret;
```
ernestask, line 762: I’m wondering… What do you think about simply using `errno`? There’s nothing that prevents us from doing so and documenting the magic `(uid_t) -1) value to mean “check errno”. 

Realistically, you would only reach the millions in the container world, but I see that at least my Fedora installation limits the subordinate IDs to 600100000 (which is well below `UINT32_MAX` and `INT32_MAX`).

Additional note is that the signedness of `uid_t` is undefined: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html, which would explain `MAX_UID_T`.

mzidek-gh:
Hi, not sure if this is a concern, but this change breaks the ABI. IIUC this is public symbol for libreport.

mgrabovsky:
> Hi, not sure if this is a concern, but this change breaks the ABI. IIUC this is public symbol for libreport.

Yeah, I don't think we've been any consistent in this regard until now, so… If I recall correctly, nobody else is using these APIs (repoquery'd Fedora repos), so we should be relatively safe.

Of course, it'd be safer to bump the version accordingly whenever a breaking change occurs.

mgrabovsky:
OK, I updated both pull requests to use `errno`. Contrary to my belief, it is already being used in the codebase.

ernestask:
Holy wow, I didn’t expect that.

PR #631, opened 2020-05-06
setgid instead of setuid the abrt-action-install-debuginfo-to-abrt-cache [RHBZ 1796245]

xsuchy:
This is called by abrt-action-install-debuginfo-to-abrt-cache which used to be setuid and now is setgid binary.
Therefore we do not need to setuid and we neither need it.

### DIFF
--- a/src/client-python/reportclient/debuginfo.py
+++ b/src/client-python/reportclient/debuginfo.py
@@ -35,41 +35,38 @@ from reportclient import (_, log1, log2, RETURN_OK, RETURN_FAILURE,
                           error_msg)
 
 
-def ensure_abrt_uid(fn):
+def ensure_abrt_gid(fn):
     """
-    Ensures that the function is called using abrt's uid and gid
+    Ensures that the function is called using abrt's gid
 
     Returns:
         Either an unchanged function object or a wrapper function object for
         the function.
     """
 
-    current_uid = os.getuid()
     current_gid = os.getgid()
     abrt = pwd.getpwnam("abrt")
 
     # if we're are already running as abrt, don't do anything
-    if abrt.pw_uid == current_uid and abrt.pw_gid == current_gid:
+    if abrt.pw_gid == current_gid:
         return fn
 
     def wrapped(*args, **kwargs):
         """
         Wrapper function around the called function.
 
-        Sets up uid and gid to match abrt's and after the function finishes
-        rolls its uid and gid back.
+        Sets up gid to match abrt's and after the function finishes
+        rolls its gid back.
 
         Returns:
             Return value of the wrapped function.
         """
 
-        # switch to abrt
+        # switch to abrt group
         os.setegid(abrt.pw_gid)
-        os.seteuid(abrt.pw_uid)
         # extract the files as abrt:abrt
         retval = fn(*args, **kwargs)
         # switch back to whatever we were
-        os.seteuid(current_uid)
         os.setegid(current_gid)
         return retval
 
@@ -79,7 +76,7 @@ def ensure_abrt_uid(fn):
 # TODO: unpack just required debuginfo and not entire rpm?
 # ..that can lead to: foo.c No such file and directory
 # files is not used...
-@ensure_abrt_uid
+@ensure_abrt_gid
 def unpack_rpm(package_full_path, files, tmp_dir, destdir, exact_files=False):
     """
     Unpacks a single rpm located in tmp_dir into destdir.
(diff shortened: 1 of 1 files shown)


xsuchy:
@ernestask I added that chmod() and according to my tests, it works in a multi-user setup.

After extracting the files to /var/cache/abrt-di I `chmod g+w` to dir. I do not do that to files because:

a) it was this way previously - eg. r--r--r-- on files
b) when removing file the ACL of a directory is used.

src/client-python/reportclient/debuginfo.py
```diff
@@ -420,6 +418,10 @@ def download(self, files, download_exact_files=False):
 
             progress_observer.downloaded_pkgs += 1
 
+        for root, dirs, files in os.walk(self.cachedir):
```
ernestask, line 421: Let’s not bloat the function any more and move this out to the caller script.

src/client-python/reportclient/debuginfo.py
```diff
@@ -123,6 +120,7 @@ def unpack_rpm(package_full_path, files, tmp_dir, destdir, exact_files=False):
 
     print(_("Caching files from {0} made from {1}").format("unpacked.cpio", os.path.basename(package_full_path)))
 
+    os.umask(0x002)
```
ernestask, line 123: Why suddenly hexadecimal and not octal?

xsuchy: that is typo.

src/client-python/reportclient/debuginfo.py
xsuchy: You mean to src/plugins/abrt-action-install-debuginfo.in in abrt package? That would break encapsulation.

xsuchy:
I believe that the umask() is needed for top-level directory in /var/cache/abrt-di/ because that os.walk do not go through them.

src/client-python/reportclient/debuginfo.py
ernestask: Yes, that’s what I mean, since it passes down the cache directory.

ernestask: I assume that’s what “breaking encapsulation” means.

src/client-python/reportclient/debuginfo.py
```diff
@@ -420,6 +418,10 @@ def download(self, files, download_exact_files=False):
 
             progress_observer.downloaded_pkgs += 1
 
+        for root, dirs, files in os.walk(self.cachedir):
+            for walked_dir in dirs:
+                os.chmod(os.path.join(root, walked_dir), 0o775)
```
ernestask, line 423: This will also not work for existing caches, right? Those will still be owned by abrt:abrt.

xsuchy: Yes. That 775 is in fact just g+w. It does not alter the ownership. The ownership should be either abrt:abrt (from past) or UNPRIV:abrt (since this change). Both way it will work.

ernestask: I’m not sure it will.

If I have /var/cache/abrt-di already populated and owned by abrt:abrt with directories being 755:

1. I won’t be able to write there
2. I won’t be able to subsequently call `chmod`

There already is a thing in the abrt spec file that recursively chowns the cache as abrt:abrt.

ernestask: https://github.com/abrt/abrt/blob/434cfd609f8e9f76175d7c3160c764d837f9b73d/abrt.spec.in#L574 is what I’m talking about here.

ernestask: I also found my old code with this pearl:

```
+                        if os.path.islink(path):
+                            # fchmodat in glibc does not support AT_SYMLINK_NOFOLLOW
+                            # and it would otherwise fail because some links
+                            # under .build-id can be broken.
+                            continue
+                        stat_ = os.stat(path)
+                        os.chmod(path, stat_.st_mode | stat.S_IWGRP)
```

Though I can’t remember why I did that. I’ll try to see if it really fails with those links.

xsuchy: Yeah, that `chown -R abrt:abrt` can be changed to `chown :abrt` - or even precisely, I think we can remove it as abrt 1 was looong ago. I wanted to do that in separate PR. Hmm, on the other hand I can replace it by `chmod -R 755`.

ernestask: Oh, right, you ignore regular files when changing permissions, so you can ignore that comment.

PR #634, opened 2020-05-13
WIP: Add Doxygen

mzidek-gh:
I would like to document all publicly exported symbols in our (public) header files in doxygen. Now this patch does not achieve that, but there are different goals here.

I added Doxyfile and included one header file there (src/include/dump_dir.h).

I would like to know what you guys thing about the format. Mostly if you have some strong feelings about the way the structures are documented the, I decided to use the "comment before symbol" way rather then after (with the additional <), I think it makes the code easier to look at, but is a bit longer, so I do not have strong opinions on that.

With this one patch you can call "doxygen" in the top github directory and it will generate doxygen_output direcotry with generated html. In order to look at something useful there navigate to Files -> file list -> dump_dir.h . This is the only useful generated file there for now. If you click on some symbol it should get you more detailed description (or TODO: Add description line :) )

IF you are overall OK with the format I will add other header files with actual documentation (not as many todo lines, which is work in progress, but I would like to avoid much rewriting, so it is excluded for now, before get some comments/green/red lights from others :))


mzidek-gh:
And I pushed wrong patches :D , rebased and fixed now hopefully.

mgrabovsky:
Overall, I'm OK with the style. I think I'd _very_ slightly prefer the shorter style for attributes, but I don't mind this one either.

By the way, there's already `doc/Doxyfile.in` in libreport, as well as `apidoc/Doxyfile.in` + `docs/doxygen.conf` in satyr, and `apidoc/Doxyfile.in` + `doc/doxygen.cfg` in abrt. (No idea why the duplicity. Seems like somebody had already had similar ideas.) I suppose it would be nice to unify the style for all our projects while we're at it.

mzidek-gh:
I completely missed the existing Doxyfile :) , I will merge the two. There are a lot of "half doxygen" comments in the libreport, so you are right, someone already had similar idea before.

ernestask:
Just putting it out there: https://github.com/abrt/libreport/pull/595 would mean transitioning to gtk-doc, as the annotations in documentation are used in binding generation.

It should have no effect on documenting things per se, though.

mzidek-gh:
I need to resolve some conflicts with the latest master. Marking this as WIP for now.

mzidek-gh:
I think I will close this PR and open a new for each header file. It will be smaller and more focused.

mzidek-gh:
Closing, see last comment.

PR #635, opened 2020-05-20
Versioning

mzidek-gh:
Hi, this part of the "we want to start using version info".

Note that we used version info already just never updated it (so not really used it).

One patch adds symbol table for libreport.
The first patch adds hint on how to use version-info (I think it is nice to have it there, but maybe it should be somewhere else).

NOTE: For some reason I can not compile libreport now, it is not related to this patches, but I need to clean up my machine before I test it, so consider this WIP


mzidek-gh:
Ah, I think I did some errors when rebasing, some parts are missing. No need to review this for now.

src/lib/Makefile.am
```diff
@@ -94,7 +106,8 @@ libreport_la_CPPFLAGS = \
     -D_GNU_SOURCE
 libreport_la_LDFLAGS = \
     -ltar \
-    -version-info 0:1:0
+    -version-info 0:1:0 \
```
ernestask, line 109: What do you think about using something like https://gitlab.gnome.org/GNOME/glib/-/blob/ac396b4dc14f53dda7c7263f3c4ae10c04e8b5d4/configure.ac#L41 instead?

mzidek-gh: It seems like more complicated way if using the version info to me. Like the way they do odd and even versions for debugging. I am not sure what is would be the benefit of that for us.

ernestask: My point was more using macros and not hardcoding versions.

mzidek-gh: Howver I miss in the patches LD version script, which, I added but somehow deleted with the po file changes (sigh). Here is info about version scripts. The .sym file is in format conusamable by LD version scripts.

Here is info about the LD scripts: https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html

src/lib/Makefile.am
```diff
@@ -14,6 +14,16 @@ lib_LTLIBRARIES = \
...
+# When releaseing new version of the libraries follow this guide on how to
+# update the version info (-version-info current:revision:age):
+#   1. If the library source code has changed at all since the last update, then increment revision (‘c:r:a’ becomes ‘c:r+1:a’).
+#   2. If any interfaces have been added, removed, or changed since the last update, increment current, and set revision to 0.
+#   3. If any interfaces have been added since the last public release, then increment age.
+#   4. If any interfaces have been removed or changed since the last public release, then set age to 0.
+#
+# Source and more info:
+# https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
+
```
xsuchy, line 26: Can you document this in some top-level document HOW-TO-RELEASE.md ?

mzidek-gh:
I did a bunch of changes to the PR (now it should build fine). I added a RELEASE_PROCESS.md (as xsuchy suggested) file with the hint on how to update the version info. In the future we can document the whole release process there.

mzidek-gh:
Btw. the failure to build that I mentioned before was related to the fact that I added the libreport.sym in the Makefile as an extra build dependency and not as a source file. Having it as source works fine. Not sure why that is the case, but I will not investigate that now.

mzidek-gh:
Just noticed that I also included the patch "git: Add some binaries to gitignore". It is not really related to this PR on the other hand it is very simple patch, so I am leaving it there for now, if you think it is better to move it out of this PR, just tell me.

src/lib/Makefile.am
mzidek-gh: I added RELEASE_PROCESS.md . It would be good to document the whole release process there in the future.

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,11 @@
...
+
+When releaseing new version of the libraries follow this guide on how to 
+update the version info (-version-info current:revision:age): 
+   1. If the library source code has changed at all since the last update, then increment revision (‘c:r:a’ becomes ‘c:r+1:a’). 
+   2. If any interfaces have been added, removed, or changed since the last update, increment current, and set revision to 0. 
+   3. If any interfaces have been added since the last public release, then increment age. 
+   4. If any interfaces have been removed or changed since the last public release, then set age to 0. 
+ 
+Source and more info: 
+* [**Libtool documentation**](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html)
```
xsuchy, line 11: Merge or link with https://github.com/abrt/abrt/wiki/Release-Process

src/lib/Makefile.am
mzidek-gh: From ernestask:
"My point was more using macros and not hardcoding versions."

Sorry, I completely missed that you wrote this. However, I do not kow what is the advantage of using such macros (there probably is some, otherwise people would not make them, but modifying 3 numbers looks easier to me and in general I would prefer to avoid creating more complicated logic that way, because at least for me this macro language is a bit hard to debug).

mzidek-gh:
Ah. I switched back to the la_DEPENDENCIES instead of the SOURCES variant, which i previously mentioned was failing to build in jenkins (and it is still the case, so it is failing to build :) ), I will switch it to SOURCES now.

mzidek-gh:
ok, I need to fix some rpmbuild errors. I will update it in a few minutes. There are some missing files, so it builds locally but not in rpmbuild. Need to add them.

mzidek-gh:
Fixed now.

src/gtk-helpers/Makefile.am
```diff
@@ -37,6 +38,11 @@ libreport_gtk_la_CPPFLAGS = \
     -D_GNU_SOURCE
 libreport_gtk_la_LDFLAGS = \
     -version-info 0:1:0
+
+if HAVE_LD_VERSION_SCRIPT
+libreport_gtk_la_LDFLAGS += -Wl,--version-script=$(top_srcdir)/src/gtk-helpers/libreport-gtk.sym
```
ernestask, line 43: Why `top_srcdir` when you have `srcdir` or even `abs_srcdir`?

mzidek-gh: No reason, really. I guess I am used to top_srcdir more :) . I did not notice that srcdir is already used in the file. I changed the top_srcdir to srcdir for consistency in the latest version.

PR #636, opened 2020-05-31
Replace various helper functions with GLib ones

michalfabik:
See commits. This is mainly to get rid of bloat and various reinvented wheels. Feel free to suggest more similar changes. See also https://github.com/abrt/abrt/pull/1479 and https://github.com/abrt/abrt-java-connector/pull/10


src/plugins/reporter-rhtsupport.c
```diff
@@ -123,33 +123,28 @@ int create_tarball(const char *tempfile, struct dump_dir *dd,
...
                 );
-                free(xml_name);
             }
         }
     }
 
     /* append all files from dump dir */
     dd_init_next_file(dd);
-    char *short_name, *full_name;
+    g_autofree char *short_name, *full_name;
```
mgrabovsky, line 139: Just noticed this: `g_autofree` requires that the variable be initialised, e.g. to `NULL`.

michalfabik: Interesting, I'd swear I remember fixing this very line, thanks.

src/plugins/reporter-rhtsupport.c
```diff
@@ -123,33 +123,29 @@ int create_tarball(const char *tempfile, struct dump_dir *dd,
...
-                free(xml_name);
             }
         }
     }
 
     /* append all files from dump dir */
     dd_init_next_file(dd);
-    char *short_name, *full_name;
+    g_autofree char *short_name = NULL;
+    g_autofree char **full_name = NULL;
```
mgrabovsky, line 140: This should only be a single pointer.

ernestask: https://wiki.c2.com/?ThreeStarProgrammer

michalfabik: Thanks, fixed.

src/gui-wizard-gtk/wizard.c
```diff
@@ -1162,7 +1162,7 @@ void update_gui_state_from_problem_data(int flags)
     }
     else
     {
-        int count = libreport_xatoi(count_str);
+        int count = g_ascii_strtoll(count_str, NULL, 10);
```
mgrabovsky, line 1163: `strtoll` converts to a 64-bit integer, which is wider than `int`. (Same with the other variations you used.)

src/lib/strbuf.c
```diff
@@ -141,8 +141,8 @@ char *libreport_strremovech(char *str, int ch)
 
 struct strbuf *libreport_strbuf_new(void)
 {
-    struct strbuf *buf = g_malloc0(sizeof(*buf));
-    /*buf->len = 0; - done by g_malloc0 */
+    struct strbuf *buf = g_new0(struct strbuf, 1);
+    /*buf->len = 0; - done by g_new0 */
```
mgrabovsky: Useless "comment".

src/lib/dump_dir.c
```diff
@@ -2430,7 +2430,7 @@ int dd_copy_file_unpack(struct dump_dir *dd, const char *name, const char *sourc
 int dd_create_archive(struct dump_dir *dd, const char *archive_name,
         const_string_vector_const_ptr_t exclude_elements, int flags)
 {
-    if (libreport_suffixcmp(archive_name, ".tar.gz") != 0)
+    if (g_str_has_suffix(archive_name, ".tar.gz") == 0)
```
mgrabovsky, line 2418: I would find `!g_str_has_suffix` slightly clearer.

src/gui-wizard-gtk/wizard.c
```diff
@@ -1966,7 +1966,7 @@ static bool check_minimal_bt_rating(const char  *event_name,
 
     if (!event_name)
         error_msg_and_die(_("Cannot check backtrace rating because of invalid event name"));
-    else if (libreport_prefixcmp(event_name, "report") != 0)
+    else if (g_str_has_prefix(event_name, "report") == 0)
```
mgrabovsky, line 1963: I would find `!g_str_has_prefix` slightly clearer. (+ in all the other occurrences.)

src/lib/event_config.c
```diff
@@ -201,27 +201,27 @@ static void load_config_files(const char *dir_path)
...
             {
                 opt = elem->data;
-                // log_warning("conf: replacing '%s' value:'%s'->'%s'", name, opt->value, value);
+                // log_warning("conf: replacing '%s' value:'%s'->'%s'", (char *)name, opt->value, (char *)value);
                 free(opt->eo_value);
             }
             else
             {
-                // log_warning("conf: new value %s='%s'", name, value);
+                // log_warning("conf: new value %s='%s'", (char *)name, (char *)value);
```
mgrabovsky, line 220: Either uncomment the log (though it seems more debug-level than a warning) or drop it altogether (same a few lines above).

src/lib/global_configuration.c
```diff
@@ -68,13 +68,13 @@ bool libreport_load_global_configuration_from_dirs(const char *dirs[], int dir_f
 
         map_string_iter_t iter;
         g_hash_table_iter_init(&iter, s_global_settings);
-        const char *key, *value;
-        while(libreport_next_map_string_iter(&iter, &key, &value))
+        gpointer key, value;
+        while(g_hash_table_iter_next(&iter, &key, &value))
```
mgrabovsky, line 72: `value` is not used so it can be `NULL`.

src/lib/dump_dir.c
```diff
@@ -1426,13 +1424,13 @@ void dd_sanitize_mode_and_owner(struct dump_dir *dd)
         error_msg_and_die("dump_dir is not opened"); /* bug */
 
     dd_init_next_file(dd);
-    char *short_name;
+    g_autofree char *short_name = NULL;
```
mgrabovsky, line 1427: Might leak.

src/lib/dump_dir.c
```diff
@@ -1647,13 +1643,13 @@ int dd_chown(struct dump_dir *dd, uid_t new_uid)
     else
     {
         dd_init_next_file(dd);
-        char *short_name;
+        g_autofree char *short_name = NULL;
```
mgrabovsky, line 1646: Might leak.

src/lib/dump_dir.c
```diff
@@ -2488,7 +2472,8 @@ int dd_create_archive(struct dump_dir *dd, const char *archive_name,
 
     /* Write data to the tarball */
     dd_init_next_file(dd);
-    char *short_name, *full_name;
+    g_autofree char *short_name = NULL;
+    g_autofree char *full_name = NULL;
     while (dd_get_next_file(dd, &short_name, &full_name))
```
mgrabovsky, line 2477: Might leak.

src/lib/problem_data.c
```diff
@@ -498,32 +498,32 @@ int problem_data_load_dump_dir_element(struct dump_dir *dd, const char *name, ch
 
 void problem_data_load_from_dump_dir(problem_data_t *problem_data, struct dump_dir *dd, char **excluding)
 {
-    char *short_name;
-    char *full_name;
+    g_autofree char *short_name = NULL;
+    g_autofree char *full_name = NULL;
```
mgrabovsky, line 502: Will also leak.

src/lib/problem_report.c
```diff
@@ -203,25 +203,24 @@ load_stream(FILE *fp)
 
     sections = g_list_append(sections, master);
 
-    char *line;
+    g_autofree char *line = NULL;
```
mgrabovsky, line 206: Also leak.

src/lib/run_event.c
```diff
@@ -152,7 +152,7 @@ GList *load_rule_list(GList *rule_list,
     /* Used only for better warning message */
     int line_counter = 0;
     /* Read and remember rules */
-    char *next_line;
+    g_autofree char *next_line = NULL;
```
mgrabovsky, line 155: Might leak, too.

src/lib/user_settings.c
```diff
@@ -123,14 +121,12 @@ GList *libreport_load_words_from_file(const char* filename)
         {
             /* every line is one word
              */
-            char *line;
+            g_autofree char *line = NULL;
             while ((line = libreport_xmalloc_fgetline(fp)) != NULL)
```
mgrabovsky, line 125: Leaky.

src/plugins/reporter-rhtsupport.c
```diff
@@ -123,33 +123,29 @@ int create_tarball(const char *tempfile, struct dump_dir *dd,
...
             }
         }
     }
 
     /* append all files from dump dir */
     dd_init_next_file(dd);
-    char *short_name, *full_name;
+    g_autofree char *short_name = NULL;
+    g_autofree char *full_name = NULL;
     while (dd_get_next_file(dd, &short_name, &full_name))
```
mgrabovsky, line 141: These will leak, too.

src/gui-wizard-gtk/wizard.c
michalfabik: But you told me to use explicit `== 0` with `strcmp` just the other day! :)

src/plugins/reporter-print.c
```diff
@@ -80,7 +80,7 @@ int main(int argc, char **argv)
         /* We used freopen to change stdout,
          * but libreport_ask() writes to stdout. Can't use that trick anymore.
          */
-        char *msg = NULL;
+        g_autofree char *msg = NULL;
```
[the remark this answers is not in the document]
michalfabik: Oh, for some reason I thought `g_autofree` takes care of freeing before reassignment too. Thanks.

src/gui-wizard-gtk/wizard.c
mgrabovsky: Well, `g_str_has_prefix` return a Boolean (`0` or `1`) while `strcmp` may return `-1`, `1` or `0`, so the semantics are slightly different. I mean, it's not a big deal, it's just my point of view on these syntactic matters.

michalfabik: Oh, good point. Never thought of that. Probably because I don't remember ever seeing the distinction between `strcmp` returning 1 and -1 put to actual use.

src/plugins/reporter-print.c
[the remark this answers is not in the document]
ernestask: The deal with `g_autofree` and friends is that, where supported (it depends on compiler extensions, but I think only MSVC does not have that implemented out the major ones), it results in cleanup being performed when the variables goes out of scope, so you would need to plaster the function with compound statements if you want to avoid ever having to manually free anything.

src/gui-wizard-gtk/wizard.c
michalfabik: So, a different function, or just cast to `int`?

mgrabovsky: It appears GLib doesn't have a function for 32-bit integers, so one option would be to load the value into a `long`, check all the possible errors (overflow, underflow, errors from `g_ascii_strtoll`) and finally cast it to `int`. The other option is to stick with `libreport_xatoi` for now. It seems to be adequate for the task at hand.

michalfabik: Like [this](https://github.com/abrt/libreport/pull/636/commits/316297054a57f81578a26af28724d30b3a22374f)? (I left the tests as they were for the time being as I'll likely remove them altogether - there's no point in testing whether `g_ascii_strtoll` works)

mgrabovsky: Well, yes, but it makes sense to wrap that up in a function and then we're back to `libreport_xatoi`…

src/lib/global_configuration.c
mgrabovsky: What I meant was that you could call `g_hash_table_iter_next(&iter, &key, NULL)`.

src/plugins/reporter-bugzilla.c
```diff
@@ -110,15 +110,22 @@ static void set_default_settings(map_string_t *osinfo, map_string_t *settings)
 static void set_settings(struct bugzilla_struct *b, map_string_t *settings)
 {
     const char *environ;
+    g_autofree gpointer login = NULL, pw = NULL, bzurl = NULL;
 
     environ = getenv("Bugzilla_Login");
-    b->b_login = g_strdup(environ ? environ : libreport_get_map_string_item_or_empty(settings, "Login"));
+    if (!environ)
+        login = g_hash_table_lookup(settings, "Login");
+    b->b_login = g_strdup(environ ? environ : (login ? login : ""));
```
mgrabovsky: This (and the related changes) seems a bit convoluted.

What about the following?

```c
login = getenv("Bugzilla_Login");
if (!login)
    login = g_hash_table_lookup(settings, "Login");
b->b_login = g_strdup(login ? : "");
```

src/lib/event_config.c
```diff
@@ -284,10 +280,10 @@ GHashTable *load_event_config_data(void)
      */
     load_config_files(EVENTS_CONF_DIR);
 
-    char *cachedir;
-    cachedir = libreport_concat_path_file(g_get_user_cache_dir(), "abrt/events");
+    g_autofree char *cachedir = NULL;
+    g_autofree char *user_cache_dir = (char*)g_get_user_cache_dir();
```
ernestask, line 284: Oh my god, no, why?

src/lib/event_config.c
```diff
@@ -284,10 +280,10 @@ GHashTable *load_event_config_data(void)
...
+    cachedir = g_build_filename(user_cache_dir ? user_cache_dir : "", "abrt/events", NULL);
```
ernestask, line 285: This makes no sense, you will never get a null string.

PR #637, opened 2020-06-02
ureport: Drop Strata integration

ernestask:
Strata is being sunset with no replacement for our used API endpoints in
sight. This commit drops the integration code, because no one pretty
much wants it, either.


ernestask:
Requesting some reviews, because this has stalled. I’ll get on the conflicts in the meantime.

michalfabik:
Nitpick: `tests/ureport/rhsm/__pycache__/*` isn't needed in `.gitignore` any more.

src/include/ureport.h
```diff
@@ -111,7 +111,7 @@ libreport_ureport_server_config_set_url(struct ureport_server_config *config,
  *
  * @param config Where the paths are stored
  * @param client_path Path in form of cert_full_path:key_full_path or one of
- *        the following string: 'rhsm', 'puppet'.
+ *        the following string: 'puppet'.
```
mgrabovsky, line 114: "or one of the following string" 🤔

ernestask: The string was ahead of its time.

ernestask:
Addressed both comments, going to proceed with the merge.

PR #638, opened 2020-06-02
internal: Remove useless #define

mgrabovsky:
This `#define` was introduced in fd2f416b3 for no obvious reason and apparently does nothing.


ernestask:
I don’t know, looks important to me.

PR #641, opened 2020-06-11
GitLab reporter

mgrabovsky:
A first approximation for the new reporter.

So far, I've only tested the CLI with a mock API server and a live instance.

Still a work-in-progress. No duplicate detection yet, doesn't check if the problem has already been reported, error handling could be improved, no attachments yet, etc.

Related: abrt/abrt#1453


mgrabovsky:
Feedback is welcome regarding the design and general principles. Implementation details might still fluctuate.

mgrabovsky:
There's a slight issue with how the report ends up formatted in GitLab, but I'm afraid there's no easy way out with the current formatting approach.

image

ernestask:
> There's a slight issue with how the report ends up formatted in GitLab, but I'm afraid there's no easy way out with the current formatting approach.

Have you tried adding backticks? Looks trivial at a glance.

mgrabovsky:
Should be ready for review now.

A few questions are left: Which components do we roll this out for? I've picked random GNOME projects for a tentative list so far. Is this something that needs to be agreed on with GNOME people?

ernestask:
It should be coordinated with GNOMies who are also Fedora packagers, because this is strictly Fedora-specific.

mgrabovsky:
Some more work will be necessary. This only allows one GitLab instance per config file while not all GNOME projects are on gitlab.gnome.org.

mgrabovsky:
Closing as there is no capacity to maintain this.

PR #642, opened 2020-06-17
Fix Coverity defects

mgrabovsky:
Fixed a couple of issues found by Coverity.

### DIFF
--- a/src/lib/make_descr.c
+++ b/src/lib/make_descr.c
@@ -81,9 +81,6 @@ char *libreport_make_description(problem_data_t *problem_data, char **names_to_s
 
     GString *buf_dsc = g_string_new(NULL);
 
-    const char *type = problem_data_get_content_or_NULL(problem_data,
-                                                            FILENAME_TYPE);
-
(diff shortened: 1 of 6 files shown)


src/lib/make_descr.c
```diff
@@ -236,6 +233,10 @@ char *libreport_make_description(problem_data_t *problem_data, char **names_to_s
 
     if (desc_flags & MAKEDESC_SHOW_MULTILINE)
     {
+        const char *type = problem_data_get_content_or_NULL(problem_data,
+                                                            FILENAME_TYPE);
+        bool is_kernel_oops = type != NULL && strcmp(type, "Kerneloops") == 0;
```
ernestask, line 238: Just as an FYI: `g_strcmp0()` handles null strings, so you could save on typing, but this is fine, too.

mgrabovsky: Actually, I'll change that. Thanks for the tip.

ernestask:
I’ll fix the failing job. Stupid fragile scripts.

PR #647, opened 2020-06-24
tests: Rewrite (de)compression test

ernestask:
Seh komm-mits.


ernestask:
BAM, renamed.

ernestask:
Where are you looking? Since it’s not part of the Autotest suite, the output is in a different place (during `make check`, it will be run before everything else and the summary will be in tests/test-suite.log).

mgrabovsky:
Ah, right, I missed it at the very top.

PR #650, opened 2020-06-30
Fix more memory management errors

mgrabovsky:
The `g_autofree` was incorrectly added in 05e9c9273. This causes a double-free error and crash.

These strings are needed elsewhere and are freed later by `msg_content_free()`.

### DIFF
--- a/src/lib/curl.c
+++ b/src/lib/curl.c
@@ -685,12 +685,14 @@ char *libreport_upload_file_ext(post_state_t *state, const char *url, const char
                 (state->curl_result == CURLE_LOGIN_DENIED
                  || state->curl_result == CURLE_REMOTE_ACCESS_DENIED))
         {
-            g_autofree char *msg = g_strdup_printf(_("Please enter user name for '%s//%s':"), scheme, hostname);
+            char *msg = g_strdup_printf(_("Please enter user name for '%s//%s':"), scheme, hostname);
             username = libreport_ask(msg);
+            free(msg);
             if (username != NULL && username[0] != '\0')
             {
                 msg = g_strdup_printf(_("Please enter password for '%s//%s@%s':"), scheme, username, hostname);
                 password = libreport_ask_password(msg);
+                free(msg);
                 /* What about empty password? */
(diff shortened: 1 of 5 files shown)


src/lib/curl.c
```diff
@@ -685,12 +685,14 @@ char *libreport_upload_file_ext(post_state_t *state, const char *url, const char
                 (state->curl_result == CURLE_LOGIN_DENIED
                  || state->curl_result == CURLE_REMOTE_ACCESS_DENIED))
         {
-            g_autofree char *msg = g_strdup_printf(_("Please enter user name for '%s//%s':"), scheme, hostname);
+            char *msg = g_strdup_printf(_("Please enter user name for '%s//%s':"), scheme, hostname);
             username = libreport_ask(msg);
```
ernestask, line 689: Unrelated, but this is kind of stupid. Should rather be a printf-like function.

src/plugins/reporter-mantisbt.c
```diff
@@ -695,6 +695,7 @@ int main(int argc, char **argv)
                     if (g_list_find_custom(ii->mii_attachments, name, (GCompareFunc) strcmp) == NULL)
                         break;
 
+                    free(name);
```
ernestask, line 698: If you’re going to free it, then also unset the pointer. Something like `g_clear_pointer(&name, free)`.

ernestask: That assumes that there exists a path where the loop will reach the line on the last iteration. Not sure how plausible that is, but something to consider.

ernestask: Right, and there is no “last iteration”, because it just goes.

mgrabovsky:
> Looks fine, thanks. I guess I’ll also do the Valgrind treatment here for CI, so we can at least find something (though getting around false positives might be annoying) before the code is merged.

Funny, I just found out there's a Make target `maintainer-check-valgrind` in `tests/` in both libreport and abrt. Looking at the results, I'm not sure if anybody has ever run it.

PR #654, opened 2020-07-22
Fix issues reported by valgrind

michalfabik:
I started resolving issues reported by valgrind when running `make maintainer-check-valgrind` in `tests`, that @mgrabovsky mentioned in another PR.

### DIFF
--- a/src/gui-wizard-gtk/wizard.c
+++ b/src/gui-wizard-gtk/wizard.c
@@ -1362,8 +1362,7 @@ static void terminate_event_chain(int flags)
     if ((flags & TERMINATE_WITH_RERUN))
         return;
 
-    free(g_event_selected);
-    g_event_selected = NULL;
+    g_clear_pointer(&g_event_selected, free);
 
     g_list_free_full(g_auto_event_list, free);
     g_auto_event_list = NULL;
@@ -2576,8 +2575,7 @@ static void add_workflow_buttons(GtkBox *box, GHashTable *workflows, GCallback f
 
 static char *setup_next_processed_event(GList **events_list)
 {
-    free(g_event_selected);
-    g_event_selected = NULL;
+    g_clear_pointer(&g_event_selected, free);
 
     char *event = get_next_processed_event(events_list);
     if (!event)
--- a/src/lib/dump_dir.c
+++ b/src/lib/dump_dir.c
@@ -1367,11 +1367,12 @@ void dd_create_basic_files(struct dump_dir *dd, uid_t uid, const char *chroot_di
     if (!dd_exist(dd, FILENAME_HOSTNAME))
         dd_save_text(dd, FILENAME_HOSTNAME, buf.nodename);
 
-    g_autofree char *release = load_text_file("/etc/os-release",
+    char *release = load_text_file("/etc/os-release",
                         DD_LOAD_TEXT_RETURN_NULL_ON_FAILURE | DD_OPEN_FOLLOW);
     if (release)
     {
         dd_save_text(dd, FILENAME_OS_INFO, release);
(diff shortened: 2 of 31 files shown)


src/lib/event_xml_parser.c
```diff
@@ -386,6 +386,7 @@ static void text(GMarkupParseContext *context,
 
                         log_info("event name:'%s'", text_copy);
                         ec_set_screen_name(ui, text_copy);
+                        free(text_copy);
                         text_copy = NULL;
```
mgrabovsky, line 390: You can use [`g_clear_pointer()`](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-clear-pointer) for all the instances of the free-then-clear pattern.

src/lib/dump_dir.c
```diff
@@ -253,7 +253,7 @@ static int read_number_from_file_at(int dir_fd, const char *filename, const char
     /* Our tools don't put trailing newline into one line files,
      * but we allow such format too:
      */
-    if (value_buf[total_read - 1] == '\n')
+    if (total_read && value_buf[total_read - 1] == '\n')
```
mgrabovsky, line 256: I'd vouch for `total_read != 0` just to be more explicit.

src/lib/get_cmdline.c
```diff
@@ -741,7 +741,7 @@ int libreport_get_mountinfo_for_mount_point(FILE *fin, struct mountinfo *mntnf,
         }
 
         size_t len = (pos_cur - pos_bck);
-        mntnf->mntnf_items[fn] = g_malloc(sizeof(char) * (len));
+        mntnf->mntnf_items[fn] = g_malloc(sizeof(char) * (len + 1));
```
mgrabovsky, line 744: Perhaps `g_malloc(len + 1)` would clearer. `sizeof(char)` is always 1.

mgrabovsky (commented):
Looking good overall.

Some of the commit messages could be more extensive. For instance “Fix Bug.add_comment method” – what was the problem and what is the fix exactly? Or ”tests: Add #include's // -Wimplicit-function-declaration”.

src/lib/get_cmdline.c
ernestask: In this case it would be a bit awkward, but, generally, something like `sizeof (*(mntnf->mntnf_items[fn]))` would be better IMO.

michalfabik: See latest commit.

src/gui-wizard-gtk/wizard.c
```diff
@@ -1362,8 +1362,7 @@ static void terminate_event_chain(int flags)
     if ((flags & TERMINATE_WITH_RERUN))
         return;
 
-    free(g_event_selected);
-    g_event_selected = NULL;
+    g_clear_pointer(&g_event_selected, g_free);
```
mgrabovsky, line 1365: The second argument should be `free` if `g_event_selected` was allocated with `malloc`. (Same with the others.)

michalfabik: Fixed in the cases of `g_event_selected`.  
`warning` at src/gui-wizard-gtk/wizard.c:2339 comes from `g_strdup_printf`, `text_copy` in src/lib/event_xml_parser.c comes from `g_strndup` and `linebuf` at src/lib/stdio_helpers.c:60 comes from `g_realloc`, so those should be fine with `g_free`, shouldn't they?

mgrabovsky: Yes, that seems correct.

mgrabovsky (approved):
Looks good.

PR #665, opened 2020-08-25
Use tito for releases

xsuchy:


.github/workflows/check.yml
```diff
@@ -32,13 +31,15 @@ jobs:
       - name: Configure build
         run: |
           chown -R builder. .
-          sudo -u builder -- ./autogen.sh
 
       - name: Build
-        run: sudo -u builder -- make --load-average=2
+        run: sudo -u builder -- tito buld --test --rpm
```
ernestask, line 36: There’s really no point in building things twice if you’re not storing the artifacts anywhere. Using Packit would solve the issue (and builds would happen in parallel and for multiple Fedora releases).

xsuchy: This should be the primary way of building. I put the other 'make' in `run test` so we can have the artifacts.

ernestask: I’m talking about the resulting RPMs.

ernestask: There’s also a typo in in that line.

xsuchy:
updated

xsuchy:
I added releasers for Fedora dist-git so we can do a release just with `tito release fedora-git`.

PR #666, opened 2020-08-25
Memory management and GLib structures

mgrabovsky:
* Use `GPtrArray` for pointer arrays in `reporter-mailx`.
* Declare GLib auto-cleanup functions for `problem_data_t`, `problem_formatter_t` and `problem_report_t`.

Broken out of #641.

### DIFF
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -583,8 +583,7 @@ static int run_event_on_dir_name_batch(
                 const char *dump_dir_name,
                 const char *event_name)
 {
-    int retval = -1;
-    problem_data_t *problem_data = NULL;
+    g_autoptr(problem_data_t) problem_data = NULL;
 
     event_config_t *config = get_event_config(event_name);
     if (config)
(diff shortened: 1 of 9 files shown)


src/cli/cli-report.c
```diff
@@ -583,19 +583,16 @@ static int run_event_on_dir_name_batch(
                 const char *dump_dir_name,
                 const char *event_name)
 {
-    int retval = -1;
-    problem_data_t *problem_data = NULL;
+    g_autoptr(problem_data_t) problem_data = NULL;
```
michalfabik, line 586: While we're at it, couldn't we get rid of the `typedef` in `problem_data.h` and just use `GHashTable` instead of `problem_data_t`? But I don't have feel very strongly about it.

michalfabik (approved):

src/cli/cli-report.c
mgrabovsky: Hm, I didn't notice that. However, although it's a `typedef`, there are multiple non-trivial functions around it (unlike the types we had eliminated before that were strictly just aliases), so I would keep it for now.

PR #681, opened 2020-11-02
wizard: Fix -Wincompatible-pointer-types

michalfabik:
See commit.

### DIFF
--- a/src/gui-wizard-gtk/wizard.c
+++ b/src/gui-wizard-gtk/wizard.c
@@ -1389,7 +1389,7 @@ static void cancel_processing(GtkLabel *status_label, const char *message, int t
     pango_attr_list_unref(list);
 }
 
-static void update_command_run_log(char *message, struct analyze_event_data *evd)
+static void update_command_run_log(const char *message, struct analyze_event_data *evd)
 {
     const bool it_is_a_dot = (message[0] == '.' && message[1] == '\0');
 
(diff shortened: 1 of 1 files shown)


mgrabovsky:
Where does this warnings pop up? I removed the `const`s in ce557c0f and noticed no warnings.

michalfabik:
I get it just by running `make`:  

    $ ./autogen.sh && ./configure && make  
    ...  
    Making all in gui-wizard-gtk  
    make[3]: Entering directory '.../libreport/src/gui-wizard-gtk'  
      CC       report_gtk-wizard.o  
    wizard.c: In function ‘start_event_run’:  
    wizard.c:1756:27: warning: assignment to ‘void (*)(const char *, void *)’ from incompatible pointer type ‘void (*)(char *, void *)’ [-Wincompatible-pointer-types]  
     1756 |     state->error_callback = run_event_gtk_error;  
          |  

(Sorry, I got the warning type wrong in the title.)

mgrabovsky:
Oh, right. In that case it should be enough to change the function signature and the type of `log_msg` can stay the same, right? (We need one fewer cast then.)

michalfabik:
See now.

mgrabovsky:
Ah, sorry, what I meant was to change the first argument of both `run_event_gtk_error()` and `update_command_run_log()` to be `const char *` as you did did in 4ec7114, but revert to `char *log_msg` on line 1400.

In the latest change, I'm slightly uncomfortable with casting a static string to a non-const `char *`.

mgrabovsky (approved):
Looks good.

PR #684, opened 2020-11-09
spec: Use autoreconf in %build instead of autoconf

mgrabovsky:
This way, when stuff is fixed in automake, the source tarball does not need to be re-generated in upstream to use the fix.

Resolves [rhbz#1893652](https://bugzilla.redhat.com/show_bug.cgi?id=1893652)

Backport of [downstream MR](https://src.fedoraproject.org/rpms/libreport/pull-request/5).

### DIFF
--- a/libreport.spec.in
+++ b/libreport.spec.in
@@ -284,7 +284,7 @@ data over ftp/scp...
 %autosetup
 
 %build
-autoconf
+autoreconf
 
 %configure \
 %if %{without bugzilla}


hroncok:
Note that the spec file used in copr by packit still has autoconf. (This is one of the reasons why I don't like sending spec changes upstream: I don't know how to do them properly.)

mgrabovsky:
In the spec used by Copr, `autoreconf` is actually run by our `autogen.sh`.

I just tried building this branch using the [Python 3.10 Mock config](https://copr.fedorainfracloud.org/coprs/g/python/python3.10/) and it went through, so I consider the bug fixed.

We're in the process of migrating to Tito, which means this file will most likely be gone soon anyway.

PR #687, opened 2020-11-25
README: Reference ABRT in the README

mzidek-gh:
Mention ABRT in the README to put libreport into some context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ directories (problem_data.h), and a set of tools that file reports.
 The library also provides an infrastructure (run_event.h, report_event.conf)
 for automatic execution of shell scripts working with dump directories.
 
+The library is used in [**ABRT (Automatic Bug Reporting Tool)**](https://abrt.readthedocs.io).
+
 
 ### Supported report destinations
 - **regular files**  : reporter-print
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,8 @@ directories (problem_data.h), and a set of tools that file reports.
 The library also provides an infrastructure (run_event.h, report_event.conf)
 for automatic execution of shell scripts working with dump directories.
 
+The library is used in [**ABRT (Automatic Bug Reporting Tool)**](https://abrt.readthedocs.io).
```
mgrabovsky, line 16: It's _Automated_ according to [our GitHub page](https://github.com/abrt/).

mzidek-gh: Hmm... looks like we have inconsistencies, but we use "Automatic" more then "Automated". (git grep 'Automatic Bug' vs git grep 'Automated Bug'). Also the RH documentation uses "Automatic": https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/ch-abrt

I do not know which one is correct.

mgrabovsky: Yeah, me neither 😅

Issue #688, opened 2020-11-26
Drop subscription-manager-rhsm from spec


mkutlak:
Libreport (**on** and **for** RHEL/CentOS 8/9) still requires `python3-subscription-manager-rhsm` but anything that used it was dropped in https://github.com/abrt/libreport/pull/637

See `libreport.spec` and `libreport.spec.in`:
https://github.com/abrt/libreport/blob/a65525dd5a35f00edc5aa3ca04ba5d55324c5c56/libreport.spec#L226

michalfabik:
What do you mean by "**on** and **for** RHEL/CentOS 8/9"? The way you're stressing it makes it look like there's an important distinction that I seem to be missing.

mkutlak:
The distinction is not important. I just wanted to point out that it is still used on RHEL/CentOS and it will be used in (for) future releases.

PR #689, opened 2020-11-26
spec: Drop subscription-manager-rhsm dependency

michalfabik:
python3-subscription-manager-rhsm isn't required since aeb7980e

Resolves:
https://github.com/abrt/libreport/issues/688

### DIFF
--- a/libreport.spec
+++ b/libreport.spec
@@ -222,9 +222,6 @@ Summary: %{name}'s micro report plugin
 BuildRequires: %{libjson_devel}
 Requires: %{name} = %{version}-%{release}
 Requires: libreport-web = %{version}-%{release}
-%if 0%{?rhel} && ! 0%{?eln}
-Requires: python3-subscription-manager-rhsm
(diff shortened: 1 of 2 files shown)


mkutlak:
It is in libreport.spec as well.
https://github.com/abrt/libreport/blob/a65525dd5a35f00edc5aa3ca04ba5d55324c5c56/libreport.spec#L226

michalfabik:
> BTW why are two spec files?  
  
Scratching my head about that too. I wasn't exactly involved in the autotools -> tito transition. Maybe others can comment on this?

xsuchy:
IMO the spec.in and gen-version is not used and can be safely removed.

mgrabovsky:
I'm planning to look into what remains to do to complete the Tito transition in abrt and libreport, as I did [in retrace server recently](https://github.com/abrt/retrace-server/commit/c02a9e7b46be938ada4a683ded49d6670872503f).

PR #691, opened 2020-12-01
BR make

xsuchy:
see individual commits

### DIFF
--- a/libreport.spec
+++ b/libreport.spec
@@ -34,6 +34,7 @@ BuildRequires: libxml2-devel
 BuildRequires: libtar-devel
 BuildRequires: intltool
 BuildRequires: libtool
+BuildRequires: make
 BuildRequires: texinfo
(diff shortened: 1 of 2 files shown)


libreport.spec
```diff
@@ -286,8 +287,6 @@ data over ftp/scp...
 %autosetup
 
 %build
-sed 's|DEF_VER=.*$$|DEF_VER='%{version}'|' -i gen-version
-#./gen-version
 ./autogen.sh
 autoconf
```
mkutlak, line 291: https://github.com/abrt/libreport/pull/684

This should be applied to the "new" spec.

xsuchy: but the spec.in is the old and the spec is the new one.

mzidek-gh:
ACK to the current version (note: the removal of spec.in was removed from this PR)

PR #692, opened 2020-12-02
Untranslatable strings

mzidek-gh:
Strings that are marked for translations should not include macros or any other compile time resolved values.


mzidek-gh:
Updated with positional arguments.

xsuchy (approved):

PR #694, opened 2021-01-05
gtk-helpers: Check for NULL before saving event config data

mgrabovsky:
Check that the configuration object for the current event is not null. This has been causing segfaults intermittently. I'm not yet sure how exactly this happens so this is kind of a hotfix.

Related: [rhbz#1909855](https://bugzilla.redhat.com/show_bug.cgi?id=1909855)


mzidek-gh:
This may not be a fix to the root cause, but it adds more debug info and makes the code more robust.

ACK

PR #696, opened 2021-01-29
[spec] Drop AnacondaRHEL workflow reference

msrb:
The support for Red Hat Customer Portal was removed in 1972853. This
forgotten reference in spec file is making the package FTBFS in RHEL now.


msrb:
I just realized that Matej is not around today, so I removed the review request :)

michalfabik:
Looks good to me.

michalfabik (approved):

msrb:
Thanks! :wink:

PR #697, opened 2021-02-01
Replace libtar with libarchive

michalfabik:
The libtar package is no longer maintained upstream. Besides, we're
already using libarchive elsewhere in the project.  
  
This commit breaks the corresponding valgrind test (`make maintainer-check-valgrind`) because calling `archive_read_free()` causes a segfault, whereas _not_ calling it causes a leak, obviously. I've tried various ways of closing/freeing the archive_read and archive_entry objects and generally shuffling code around but couldn't find anything that works properly. It's only happening in the test itself so it isn't really an issue. If you feel like digging around in it, I'd love to see what the solution is. I'm probably just being sloppy but I suppose a bug in libarchive isn't out of the question either.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1917456

### DIFF
--- a/src/lib/dump_dir.c
+++ b/src/lib/dump_dir.c
@@ -17,7 +17,8 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 #include <sys/utsname.h>
-#include <libtar.h>
+#include <archive.h>
+#include <archive_entry.h>
 #include <glib-unix.h>
 #include "internal_libreport.h"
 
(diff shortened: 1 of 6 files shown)


mgrabovsky (changes requested):
The build in Copr fails on `/usr/bin/ld: cannot find -ltar`. I think you need to remove line 95 in `src/lib/Makefile.am` to make it go through.

src/lib/dump_dir.c
```diff
@@ -2425,110 +2426,130 @@ int dd_create_archive(struct dump_dir *dd, const char *archive_name,
...
-    }
 
-    /* ...and check that gzip child finished successfully */
-    int status;
-    libreport_safe_waitpid(child, &status, 0);
-    if (status != 0)
+    archive_entry_free(entry);
+    archive_write_close(a);
+    archive_write_free(a);
+    r = chmod(archive_name, 0600);
```
mgrabovsky, line 2548: I'm thinking, for the sake of security, we should call `open()` with these permissions (as in the previous version) and then `archive_write_open_fd()`. Otherwise the file might be readable before the archive is finalised.

michalfabik: > I'm thinking, for the sake of security, we should call `open()` with these permissions (as in the previous version) and then `archive_write_open_fd()`. Otherwise the file might be readable before the archive is finalised.  
  
Good point.  
  
I'm not sure about the `close()`.

src/lib/dump_dir.c
```diff
@@ -2425,111 +2426,135 @@ int dd_create_archive(struct dump_dir *dd, const char *archive_name,
...
     {
-        result = -errno;
-        log_warning(_("Failed to open TAR writer"));
-        goto finito;
+        result = -archive_errno(a);
+        log_warning(_("%s"), archive_error_string(a));
+        return result;
+    }
+
+    archive_fd = open(archive_name, O_CREAT|O_WRONLY, S_IRUSR|S_IWUSR);
```
mgrabovsky, line 2471: I think `O_EXCL` would be nice here as well.

mgrabovsky, line 2471: I think O_EXCL would be nice here as well.

mgrabovsky (commented):
Looks good now.

mgrabovsky (approved):
Looks good now.

mgrabovsky:
@michalfabik Yep, and just as I hit the button, I remembered: Could you please add a changelog entry for the change? Removing a dependency might be significant for somebody.

PR #705, opened 2021-03-10
tito: Clean up tito setup

michalfabik:
Some finishing touches to complete the preparation for releasing with
tito (and to clean up the mess I made experimenting).


mgrabovsky:
When I run `tito build --test --rpm`, it seems to be using version 2.15.0, e.g. `/tmp/tito/x86_64/libreport-2.15.0-1.git.2253.6bb43a3.fc33.x86_64.rpm`.

michalfabik:
Ah, I forgot to revert the tito commit from the botched release.

mgrabovsky (changes requested):
Hm, there are still some weird versions floating around.

I'm trying this in a fresh clone of the repo (on this branch): tag a new version, e.g. 3.0.0, and run the build. The version is now correct in the file names and RPMs, but I can see that some the autotools stuff are still using 2.14.0, for example:
```
[...]
             echo '# Signature of the current package.' && \
             echo 'm4_define([AT_PACKAGE_NAME],' && \
             echo '  [libreport])' && \
             echo 'm4_define([AT_PACKAGE_TARNAME],' && \
             echo '  [libreport])' && \
             echo 'm4_define([AT_PACKAGE_VERSION],' && \
             echo '  [2.14.0])' && \
             echo 'm4_define([AT_PACKAGE_STRING],' && \
             echo '  [libreport 2.14.0])' && \
             echo 'm4_define([AT_PACKAGE_BUGREPORT],' && \
             echo '  [crash-catcher@fedorahosted.org])'; \ 
             echo 'm4_define([AT_PACKAGE_URL],' && \
             echo '  [])'; \
[...]
============================================================================
Testsuite summary for libreport 2.14.0
============================================================================
[...]
```

[In `configure.ac`](https://github.com/abrt/libreport/blob/a42d47473fe565fa3a83d0eaa3dde79fbb01564b/configure.ac#L22) we're reading the version from `libreport-version` but that file doesn't seem to be getting generated during the build.

michalfabik:
It's coming from the `DEF_VER` in gen-version`. The var used to be updated by the `release-*` make targets, IIRC. I can't seem to reproduce what you're describing but I suppose we need to get the version from somewhere in case the first test in `gen-version` fails (e.g. after unpacking the tarball). So, either the .spec or `.tito/package/libreport`?

mgrabovsky:
Yeah, I think we should get rid of `gen-version` and `libreport-version` and read the version in `configure.ac` from one of those you mention. However, we would then always get the version of the latest release even if new commits appeared in the meantime. I'm still not sure what the correct way to go about this is.

michalfabik:
Ok, it seems to pretty much work now, except:  
  
`make build --tgz --test` produces a file such as `libreport-git-1.169359d.tar.gz`, without the version (not sure if that's a bug or a feature).  
  
Otherwise, `make build --tgz` correctly produces `libreport-2.15.0.tar.gz` (from the latest upstream tag) and `make build --rpm --test` produces files such as `libreport-2.15.0-1.git.1.169359d.fc33.x86_64.rpm` (latest upstream tags plus whatever commits since then).

gen-version
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DEF_VER=2.14.0
+DEF_VER=$(grep '^Version:\s\+[0-9]\+\.[0-9]\+\.[0-9]\+$' libreport.spec | sed 's/.*\s\+//')
```
mgrabovsky, line 3: May I suggest a simpler `sed -ne '/^Version:/s/^Version:\s*//p' libreport.spec`?

mgrabovsky:
Getting the version this way looks fine to me.

I think the missing version on the HEAD tarball might be a feature.

However, running just `tito build --tgz` on this branch crashes on me:
```
$ tito build --tgz
Creating output directory: /tmp/tito
Checking for tag [2.14.0] in git repo [https://github.com/abrt/libreport.git]
Building package [2.14.0]
ERROR: Error running command: git archive --format=tar --prefix=libreport-2.14.0/ 114a3286d72e28be797a2d091c74f428044ec0d5:None --output=/tmp/tito/rpmbuild-libreportsr4fxycl/SOURCES/libreport-2.14.0.tar.initial

Status code: 128

Command output: fatal: not a valid object name: 114a3286d72e28be797a2d091c74f428044ec0d5:None

Traceback (most recent call last):
  File "/usr/bin/tito", line 33, in <module>
    sys.exit(load_entry_point('tito==0.6.16', 'console_scripts', 'tito')())
  File "/usr/lib/python3.9/site-packages/tito/cli.py", line 902, in main
    CLI().main(sys.argv[1:])
  File "/usr/lib/python3.9/site-packages/tito/cli.py", line 209, in main
    return module.main(argv)
  File "/usr/lib/python3.9/site-packages/tito/cli.py", line 384, in main
    return builder.run(self.options)
  File "/usr/lib/python3.9/site-packages/tito/builder/main.py", line 151, in run
    self.tgz()
  File "/usr/lib/python3.9/site-packages/tito/builder/main.py", line 537, in tgz
    self._setup_sources()
  File "/usr/lib/python3.9/site-packages/tito/builder/main.py", line 570, in _setup_sources
    create_tgz(self.git_root, self.tgz_dir, self.git_commit_id,
  File "/usr/lib/python3.9/site-packages/tito/common.py", line 873, in create_tgz
    run_command(git_archive_cmd)
  File "/usr/lib/python3.9/site-packages/tito/common.py", line 305, in run_command
    raise RunCommandException(command, status, output)
tito.exception.RunCommandException: Error running command: git archive --format=tar --prefix=libreport-2.14.0/ 114a3286d72e28be797a2d091c74f428044ec0d5:None --output=/tmp/tito/rpmbuild-libreportsr4fxycl/SOURCES/libreport-2.14.0.tar.initial
```

The `None` in `114a3286d72e28be797a2d091c74f428044ec0d5:None` looks like a Tito bug… I suppose? And the same thing happens in the main branch as well. (It doesn't happen in any other of our Tito-enabled projects, though.)

michalfabik:
from `tito build --tgz --debug`:  
  
Got package metadata: ['fatal:', 'path', "'.tito/packages/libreport'", 'exists', 'on', 'disk,', 'but', 'not', 'in', "'114a3286d72e28be797a2d091c74f428044ec0d5'"]  
  
I suppose this is what does causes it. The file was only created by the first ever run of `tito tag` (which was my botched attempt at the 2.15.0 release a few days ago). Now `tito build` is looking back to the last tag, not knowing that it wasn't created by tito and can't find the file there. I suppose it should work ok once I tag this release.

mgrabovsky (approved):
OK, that makes sense. Let's roll with it and see what happens then.

PR #707, opened 2021-03-24
spec: Sync upstream with distgit

michalfabik:
Since the current setup with GitHub Actions automatically opens PRs
against the distgit with each realease tag and since the spec file is
included in the release commit by tito, we want the upstream and distgit
version of the spec file to be as close as possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,7 +107,6 @@ testsuite.log
 # misc
 \#*
 *~
-# this is obsolete since moving to tito, but leaving here for some time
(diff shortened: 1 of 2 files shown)


msrb:
> the current setup with GitHub Actions automatically opens PRs
against the distgit with each realease tag

I think this should be `s/GitHub Actions/Packit/g` ?

msrb:
LGTM :wink:

msrb (approved):

PR #709, opened 2021-03-27
Add support for excluding whole elements from search for sensitive words

msrb:
Sections like "cpuinfo" or "open_fds" should be fairly safe
to ignore. There shouldn't be any URLs, tokens or passwords
in them.

Users can still review those sections and custom user searches will still
highlight words in them. We will just not flag any forbidden words in
them by default.

If needed, more sections can be easily added to the list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - ignored_words: add more "key" variations
+- Add support for excluding whole elements from search for sensitive words
 
 ## [2.15.0] - 2020-03-03
 ### Added
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -44,6 +44,7 @@ MAN5_TXT = \
 	mantisbt_format_analyzer_libreport.conf.txt \
 	mantisbt_formatdup_analyzer_libreport.conf.txt \
 	ignored_words.conf.txt \
+	ignored_elements.conf.txt \
 	forbidden_words.conf.txt \
 	libreport.conf.txt \
 	mailx.conf.txt \
--- a/doc/forbidden_words.conf.txt
+++ b/doc/forbidden_words.conf.txt
@@ -35,7 +35,7 @@ $XDG_CONFIG_HOME/abrt/settings/forbidden_words.conf
 
 SEE ALSO
 --------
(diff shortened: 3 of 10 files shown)


msrb:
This is related to https://github.com/abrt/abrt/issues/1399.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.16.0]
```
michalfabik, line 7: Please make it `[Unreleased]`, as we still don't know what the next release will be.

msrb: Good catch. I changed the version to `[Unreleased]` :wink:

doc/ignored_sections.conf.txt
```diff
@@ -0,0 +1,47 @@
...
+NAME
+----
+ignored_sections.conf - configuration file for libreport.
+
+DESCRIPTION
+-----------
+This configuration file specifies which sections should *NOT* be checked
+for sensitive data during the process of reporting a problem. This means
+that, for the sections specified in this file, no words will be highlighted
+in the User Interface (UI).
```
mgrabovsky, line 13: I think it would be sensible to mention `report-gtk` explicitly here.

mgrabovsky (commented):
Looks good overall.

We could also have the list as an entry in a configuration file (such as `libreport.conf`) instead since it will tend to be quite short, but this should works just as well.

msrb:
Meh, closed by accident :)

msrb:
> I think I’d prefer “element” or “item” instead of “section” since those are the terms we use in libreport to refer to these files (albeit somewhat inconsistently), see for instance the man page for libreport.conf.

Good point. `s/section/element/g` it is.

mgrabovsky:
Strange, after `make rpm`, `dnf upgrade` etc., I don't see the new config file listed in the RPM:

    $ rpm -qc libreport
    /etc/libreport/forbidden_words.conf
    /etc/libreport/ignored_words.conf
    /etc/libreport/libreport.conf
    /etc/libreport/report_event.conf

The man page isn't there either and `report-gtk` picks up keywords in the ignored elements even if I create the file manually.

mgrabovsky (approved):
Looks fine to me.

PR #712, opened 2021-04-14
cli: Address of local auto-variable assigned to a function parameter

mzidek-gh:
Error: CPPCHECK_WARNING (CWE-562): [#def7]
libreport-2.14.0/src/cli/cli-report.c:771: error[autoVariables]: Address of local auto-variable assigned to a function parameter.

### DIFF
--- a/src/cli/cli-report.c
+++ b/src/cli/cli-report.c
@@ -738,16 +738,16 @@ static int run_event_chain_real(struct run_event_state *run_state,
                                 GList                  *chain,
                                 int                     interactive)
 {
-    struct logging_state l_state;
+    struct logging_state *l_state = g_malloc0(sizeof(struct logging_state));
     int retval = 0;
 
     /* Blergh. */
     run_state->logging_callback = do_log2;
-    run_state->logging_param = &l_state;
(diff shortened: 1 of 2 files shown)


src/cli/cli-report.c
```diff
@@ -738,16 +738,16 @@ static int run_event_chain_real(struct run_event_state *run_state,
...
                                 int                     interactive)
 {
-    struct logging_state l_state;
+    g_autofree struct logging_state *l_state = g_malloc0(sizeof(struct logging_state));
     int retval = 0;
 
     /* Blergh. */
     run_state->logging_callback = do_log2;
-    run_state->logging_param = &l_state;
+    run_state->logging_param = l_state;
```
mgrabovsky, line 746: Doesn't this lead to a dangling pointer once `l_state` is autofreed?

mzidek-gh: Good catch :D , that would be nasty use-after-free. It should be freed in the free_run_event_state not by autofree.

mzidek-gh:
Updated.

michalfabik (approved):

PR #714, opened 2021-04-14
gui-wizard-gtk: Improve docs and add missing free

mzidek-gh:
Add missing g_strfreev. Reported by covscan. Also
add doxygen comment to the function that allocated
the resources to mention that the returned value
should be freed.

### DIFF
--- a/src/include/global_configuration.h
+++ b/src/include/global_configuration.h
@@ -32,6 +32,15 @@ bool libreport_load_global_configuration_from_dirs(const char *dirs[], int dir_f
 
 void libreport_free_global_configuration(void);
 
+/**
+ * Returns list of excluded elements
+ *
+ * This function returns list of elements in problem directory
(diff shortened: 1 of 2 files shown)


src/include/global_configuration.h
```diff
@@ -32,6 +32,15 @@ bool libreport_load_global_configuration_from_dirs(const char *dirs[], int dir_f
 
 void libreport_free_global_configuration(void);
 
+/**
+ * Returns list of excluded elements
+ *
+ * This function returns list of elements in problem directory
+ * that are always ecluded from the report.
```
mgrabovsky, line 39: Typo
```suggestion
 * that are always excluded from the report.
```

mzidek-gh: Good catch, updated.

michalfabik (approved):

PR #715, opened 2021-04-14
gui-wizard-gtk: Check if EXCLUDE_FROM_REPORT is set

mzidek-gh:
Hypothetically the g_setenv function can fail. We should
probably not continue execution in that case because users
might accidentally include data in reports that are explicitely
configured not to be included in reports.

Reported by covscan.


mzidek-gh:
Note that there are more cases where we do not check the return value of g_setenv, but I decided to fix this one because I think it might have some security implications.

michalfabik (approved):

PR #716, opened 2021-04-14
gui-wizard-gtk: Possible double free

mzidek-gh:
g_free does not need to be checked for NULL
but the pointer is not automatically set to NULL
after g_free call.

Found by covscan.

### DIFF
--- a/src/gui-wizard-gtk/wizard.c
+++ b/src/gui-wizard-gtk/wizard.c
@@ -2101,8 +2101,7 @@ static GList *find_words_in_text_buffer(int page,
 
             if (g_list_find_custom(ignore_sitem_list, found_word, (GCompareFunc)sitem_contains))
             {
-                if (found_word)
-                    free(found_word);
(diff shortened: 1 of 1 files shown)


src/gui-wizard-gtk/wizard.c
```diff
@@ -2101,8 +2101,8 @@ static GList *find_words_in_text_buffer(int page,
 
             if (g_list_find_custom(ignore_sitem_list, found_word, (GCompareFunc)sitem_contains))
             {
-                if (found_word)
-                    free(found_word);
+                g_free(found_word);
+                found_word = NULL;
```
mgrabovsky, line 2105: This would be even shorter using `g_clear_pointer()`.

mzidek-gh: True. I will update this pattern in the other PRs as well.

michalfabik (approved):

PR #718, opened 2021-04-14
dirsize: No need to check for NULL

mzidek-gh:
Reported by covscan. If we check for NULL the static
analyzers assume the value can be NULL, but it can not
as it is allocated using the g_* family of functions
and the application would terminate if the memory allocation
failed.


mzidek-gh:
This is not only to make covscan happy, I think the code is more readable this way.

michalfabik (approved):

PR #719, opened 2021-04-14
dump_dir: Use g_free instead of free

mzidek-gh:
We wrongly called free instead of g_free and we
manipulated with already freed pointer without
initializing it back to NULL.

Reported by covscan.


mzidek-gh:
Again, covscan assumes possible double-free if there is not "back to NULL" after free.

michalfabik (approved):

PR #722, opened 2021-04-14
ureport: Strange usage of tmp variable

mzidek-gh:
I do not know why this tmp variable was used this way.
Probably a leftover from some refactoring.

Again this was found by covscan.

### DIFF
--- a/src/lib/ureport.c
+++ b/src/lib/ureport.c
@@ -126,28 +126,30 @@ ureport_server_config_load_basic_auth(struct ureport_server_config *config,
 
     g_autoptr(GHashTable) settings = NULL;
 
-    g_autofree char *tmp_password = NULL;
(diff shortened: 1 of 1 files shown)


michalfabik:
recheck

michalfabik:
This breaks the `ureport_server_config_load_basic_auth` test (no. 53 in the testsuite): `munmap_chunk(): invalid pointer`.  
  
It seems to be trying to `free` something that wasn't `malloc`'d.  
  
I think the idea behind the tmp vars was to ensure proper functioning of `g_autofree` in conjunction with with the `const` pointers and `strchr`.

mzidek-gh:
@michalfabik thanks for the review. I have split the password variable into two new variables with suffixes that show where they are from. One is freed the other one is not.

PR #725, opened 2021-05-26
Use GLib for computing SHA-1 digests

mgrabovsky:
In d605ffea, we ported the generation of digests from a bundled library to Nettle. However, Red Hat security policies require that no package depend on Nettle directly. Since GLib provides functions to compute SHA-1 digests as well, we are migrating the code to use it as of now.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1956871


t184256:
Same as with [satyr](https://github.com/abrt/satyr/pull/328#issuecomment-849000948), glib is an even worse choice than nettle cert-wise, sorry.

mgrabovsky:
> This is a good opportunity to get rid of SHA1 from libreport. Please use G_CHECKSUM_SHA256 instead of the G_CHECKSUM_SHA1. It is OK if you add it as another commit in this PR.

OK, that might be a way. Are we 100% sure it doesn't break anything, though?

mzidek-gh:
Sorry, forgot to respond. We are not 100% sure it will not break anything, but the chances are small. The only thing that relies on hash values is the duplicate hash calculation so we lose duplicates detection between old and new bugs, but that is something we already talked about in the past that it is not a big deal.

mzidek-gh:
Sorry, I am dropping the request to change the hash function for now. Let's merge and think what function is most suitable later.

mzidek-gh:
ACK.

mzidek-gh (approved):

PR #729, opened 2021-08-18
Avoid direct use of libproxy

mcatanzaro:
GProxyResolver has been around since GLib 2.26. There's no reason to use
libproxy directly anymore. GProxyResolver will call into libproxy if
GLib thinks it's a good idea to do so, except in distributions that are
configured to build glib-networking without libproxy support.

Note that the existing code in get_proxy_list() was unsafely using
g_autofree and g_free() to free the memory allocated by libproxy. With
the switch to GProxyResolver, the memory is actually allocated by GLib
now, so this is now actually correct to do.

**Note that this commit is untested.** I just checked to make sure that it builds. **It will probably work, but please sanity-check to ensure it doesn't crash at runtime before merging.** The APIs are basically 1:1 except for some details around SOCKS proxies that libreport doesn't care about, so if it doesn't crash it will be good.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -224,9 +224,6 @@ PKG_CHECK_MODULES([GIO], [
 PKG_CHECK_MODULES([GOBJECT], [gobject-2.0])
 PKG_CHECK_MODULES([LIBXML], [libxml-2.0])
 PKG_CHECK_MODULES([CURL], [libcurl])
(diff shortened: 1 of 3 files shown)


michalfabik (approved):

michalfabik:
Thanks! I've tested `reporter-bugzilla` and it seems to be working fine.  
  
LGTM.

PR #730, opened 2021-08-24
spec: Drop libproxy dependency

mgrabovsky:
Dependency on libproxy was factually removed in 11df7ffc. This change follows up by removing the formal dependency in the spec and pkgconfig files.


michalfabik (approved):

michalfabik:
Thanks, I remember thinking of this when I was going through M. Catanzaro's PR but forgot about it immediately.

PR #732, opened 2021-08-24
Add reporter-bugzilla.py

michalfabik:
Still work in progress but feedback welcome. Can upload bug and attachments but not much else.  
  
Sample bug: https://bugzilla.redhat.com/show_bug.cgi?id=2077594  
  
Current most pressing issue: Whiteboard should look like "abrt_hash:87a26b99ba9fed0efdb880358dfb2324e3d3268d;VARIANT_ID=workstation" (and it does work with my local little BZ instance). Why does RHBZ strip it to the hash only? Maybe escape the (semi)colons?


msrb:
Good job so far :wink: :+1: 

I'd probably move these python files to a different directory.

What about:
`src/python/`
or:
`src/python/plugins/`

WDYT?

michalfabik:
@msrb, @mgrabovsky, PTAL. This isn't finished but I guess I'm at a stage where I could use a few more sets of eyes.  
  
What's finished:  
- utility code, i.e. all files in `src/python/plugins`other than `reporter-bugzilla.py`  
  
What actually works:  
- `reporter-bugzilla.py --help` :P  
- searching for duplicates with `reporter-bugzilla.py -h HASH`  
  
What's done but broken as of this writing:  
- `reporter-bugzilla.py -t BUG_ID FILE`
- `reporter-bugzilla.py -t BUG_ID -w`  
Both broken due to malformed request to BZ - you can add debug prints to the respective methods in `bz_connection.py` to see the responses. This will most likely require some trial and error because I can't seem to find any answers in BZ REST API documentation.  
- bug creation with `reporter-bugzilla.py -d DUMP_DIR - This actually did work but should be assumed broken ATM because I've done far too many changes since I last tested it.  
  
What I'd like you to focus on:  
- the overall spaghettiness of the code. Can we afford to simplify some methods or merge/drop some modules at the cost of reduced robustness? I dare say yes (Python should handle a lot of the low-level stuff by itself) but I'm having trouble assessing what specifically.
- The abominable exception/errno error handling in `dump_dir.py`. This is a result of my attempt to translate the C errno system into Python exception but there's not a 1:1 mapping between the too. A lot of the file can probably be done away with (see above).
- The logging system. Currently, I've got a class in each module just to be able to instantiate it in the main program and pass the common logger to it. This is quite ugly, of course.  
- any `TODOs` in the code - some are left over from the C code, some are added by me.  
- Real-life testing. 
  
What I have no clue about:  
- unit testing. Can we make use of the existing C test suite in any way?  
- integration of the python code in the libreport rpm package.   
  
I will try to finish the main program now, which should be pretty straightforward, save for any issues with the BZ requests that I've mentioned above.

michalfabik:
Added the `--debug` cmdline optoin and fixed a few errors in dump dir locking just now.

michalfabik:
Done. Will start testing now.

mzidek-gh:
We had a short pokec about this PR today with mfabik. There are are still some open issues with this PR.
1) It would be good to cover the new code with tests. I think we could call the bugzilla_reporter.py directly with some parameters and mock the response from python requests (for example using requests-mock or something else). We should cover at least the functionality that is required by src/plugins/bugzilla_event.conf and src/plugins/bugzilla_anaconda_event.conf

2) The dump_dir.py is direct rewrite from the C version, but it also uses the same method to lock the directory/files, using the symlink creation that is atomic in C, but not in Python, meaning the locking is not reliable in the Python version. I think we should either continue using the C version of the dump_dir and just add Python bindings to it or use a different locking mechanism in the Python version and ditch the C version completely.

Given the fact that none of the new code is currently used I think the above is not a blocker for merging this PR, so we have at least some code to test and play with, but we should solve the issues before we start using the code in production.

msrb:
I wouldn't worry about directory locking too much. When people call bugzilla-reporter, the content of the problem directory is usually final, i.e. all processing has been done already. And bugzilla-reporter is 99% read operations.

michalfabik:
Done. Tested build and installation on F36 and Rawhide.  
  
There have been many changes in the meantime and commits have been reorganized/squashed so the above conversation is no longer relevant. Leaving it there for historical context.  
  
Next step is a new libreport release but please have a good look at it before that.

msrb (approved):
LGTM. Great work on this -- thanks! This is huge :)

PR #733, opened 2021-09-08
reporter-bugzilla: fix subcomponent handling for RHEL

msrb:
Most of the components in Bugzilla don't have any subcomponents, so we need to be careful when handling them.


msrb:
@michalfabik PTAL :wink:

michalfabik:
Of course makes sense, LGTM.

michalfabik (approved):

PR #737, opened 2021-11-04
tests: Don't use deprecated assertEquals

michalfabik:
unittest.assertEquals was deprecated in Python 3.2 and will no longer be
supported in 3.11

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=2019402


mgrabovsky (approved):
Looks great. Thanks.

msrb:
Nice! This was easier than I thought  :+1: :wink:

PR #738, opened 2021-11-30
debuginfo-install is a dnf command

ckujau:
```
$ debuginfo-install
bash: debuginfo-install: command not found
```


michalfabik:
Good catch, thanks!

michalfabik (approved):

Issue #706, opened 2021-03-16
Is it possible to send problem reports without a conventional username+password bugzilla account?


whereswaldon:
I'm a Red Hat Associate, and I'd like to submit bug reports when applications crash to the Red Hat bugzilla instance. However, my bugzilla account uses Red Hat SSO instead of a conventional username + password combination. I haven't been able to figure out how to set a password on my bugzilla account (I suspect it's impossible by design for SSO accounts), so I'm unable to report bugs with libreport.

Is there a known workaround for this? If not, is this a feature request that would be accepted, or is it too niche?

Thanks!

mgrabovsky:
It is only possible to report bugs using a username and password combination at the moment and there are no plans to support other authentication methods.

I can change my Bugzilla password on the “Account Information” settings tab. If you can't see that in your web interface, perhaps the “Forgot Password” function in the login box might be an option.

(For what it's worth, the Bugzilla API also seems to accept API keys as an authentication token, but we would need to see a clear demand to implement this in the reporting plugin.)

whereswaldon:
Thanks for the quick response! I have attempted both changing my password in account settings (fails because I don't know my current password, as the password set for me in the DB is probably random garbage. That's how SSO accounts are normally configured on services like this), and I've tried the forgot password link (asks for my email, then redirects me and fails to send me a reset link).

I think I'm stuck unless a BZ admin helps me reset the password or using API keys is an option, but I totally understand your reservation around adding the API keys without a clearer use-case.

msrb:
I think this is definitely worth considering as we can probably expect more and more people to rely on SSO for bugzilla. It is pretty convenient after all.

msrb:
With libreport 2.17.1 it is now possible to authenticate using API key! You can generate your API key directly in Bugzilla: https://bugzilla.redhat.com/userprefs.cgi?tab=apikey :wink:

PR #742, opened 2022-02-21
[reporter-bugzilla] Use API key for authentication

msrb:
Following change will take effect on February 28 2022:

"""
Applications making API calls to Bugzilla may no longer
authenticate using passwords or supplying API keys
in call parameters. Instead, API keys must be supplied
in the Authorization header.
"""

See the announcement for more details:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/U3PQNCLKA3R6GA2UI2EZP2OT4VAKVHBI/


msrb:
The code was accidentally tested here: https://bugzilla.redhat.com/show_bug.cgi?id=2056422

msrb:
This is how the change looks like in UI:

Screenshot from 2022-02-21 08-46-44

msrb:
@michalfabik @mgrabovsky PTAL :wink: Thanks!

src/plugins/rhbz.c
```diff
@@ -208,6 +208,21 @@ bool rhbz_login(struct abrt_xmlrpc *ax, const char *login, const char *password)
     return true;
 }
 
+
+bool add_session_api_key(struct abrt_xmlrpc *ax, const char *api_key)
```
mgrabovsky, line 212: This should have the `rhbz_` prefix to be consistent with the rest of the module.

src/plugins/reporter-bugzilla.c
```diff
@@ -76,6 +76,7 @@ int attach_file_item(struct abrt_xmlrpc *ax, const char *bug_id,
 struct bugzilla_struct {
     char *b_login;
     char *b_password;
```
mgrabovsky, line 78: Is there any reason to keep these two around?

mgrabovsky (changes requested):
Looks good overall. Seems to be working quite well. Just two small suggestions.

Also: Resolves #706

src/plugins/rhbz.c
msrb: Good catch. I will fix that.

src/plugins/reporter-bugzilla.c
msrb: I think we could drop it, but I was not 100% sure and I did not want to go full ballistic and remove a lot of code just yet. I will add a comment there that the field are no longer used and that we will drop them in future release.

mgrabovsky: I think we can remove them right away as this API should not be public and it's not even versioned. I believe it's only used by `reporter-bugzilla` and nowhere else.

msrb: Actually, looking at the code, `b_login` is still used when people call `reported-bugzilla` with -w option ("Add bugzilla user to CC list"). But `b_password` is indeed a dead weight. Removing it...

src/plugins/rhbz.c
msrb: Updated.

mgrabovsky (approved):
Looks good.

msrb:
Thanks! Merging...

PR #746, opened 2022-03-10
reporter-bugzilla: send API key in HTTP header

msrb:
But only if the target instance is Red Hat Bugzilla.

This change is required as Red Hat Bugzilla expects the API key
to be in the HTTP header and it won't find it in XML-RPC params.

xmlrpc-c doesn't support custom headers. Only User-Agent header
is supported. However, since HTTP is a text protocol, we can add
more headers by simply strategically placing line break characters
("\r\n" for HTTP) inside the User-Agent header.
HTTP will do the right thing and interpret such string as multiple
headers.

### DIFF
--- a/src/lib/abrt_xmlrpc.c
+++ b/src/lib/abrt_xmlrpc.c
@@ -39,6 +39,14 @@ void abrt_xmlrpc_error(xmlrpc_env *env)
 
 struct abrt_xmlrpc *abrt_xmlrpc_new_client(const char *url, int ssl_verify)
 {
+    /* non-RH clients are the same as RH clients,
+     * but they don't send API key in the HTTP header.
+     */
+    return abrt_xmlrpc_new_redhat_client(url, ssl_verify, NULL);
+}
+
+struct abrt_xmlrpc *abrt_xmlrpc_new_redhat_client(const char *url, int ssl_verify, const char *api_key)
+{
     GList *proxies = NULL;
     xmlrpc_env env;
     xmlrpc_env_init(&env);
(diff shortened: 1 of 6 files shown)


msrb:
Tested end-to-end here: https://bugzilla.redhat.com/show_bug.cgi?id=2062659

msrb:
@mgrabovsky @michalfabik @mzidek-gh PTAL :wink:

src/lib/abrt_xmlrpc.c
```diff
@@ -70,6 +78,19 @@ struct abrt_xmlrpc *abrt_xmlrpc_new_client(const char *url, int ssl_verify)
...
+
+        /* The User-Agent header normally has the following format:
+         * User-Agent: libreport/2.15.2 Xmlrpc-c/1.51.0 Curl/7.79.1
+         *
+         * We modify it here so it becomes 3 headers:
+         * User-Agent: libreport/2.15.2
+         * Authorization: Bearer <api-key>
+         * X-Libreport-Extra-User-Agent: Xmlrpc-c/1.51.0 Curl/7.79.1
+         */
+        curl_parms.user_agent = g_strdup_printf("%s\r\nAuthorization: Bearer %s\r\nX-Libreport-Extra-User-Agent:", curl_parms.user_agent, api_key);
```
mgrabovsky, line 92: I don't think this string is ever freed, but it should be given that it contains sensitive information.

src/lib/libreport-web.sym
```diff
@@ -44,6 +44,7 @@ global:
     abrt_xmlrpc_params_set_value_str;
     abrt_xmlrpc_params_set_value;
     abrt_xmlrpc_new_client;
+    abrt_xmlrpc_new_redhat_client;
```
mgrabovsky, line 47: We should also update `-version-info` for libreport-web in `src/lib/Makefile.am`.

src/plugins/reporter-bugzilla.c
```diff
@@ -217,6 +217,17 @@ char *ask_bz_api_key(const char *message)
     return api_key;
 }
 
+
+static
+bool is_redhat_bugzilla(const char *bugzilla_url)
+{
+    if (g_str_has_suffix(bugzilla_url, "redhat.com")) {
```
mgrabovsky, line 224: I suppose this could be just `return g_str_has_suffix(bugzilla_url, "redhat.com")` :slightly_smiling_face:

msrb: Ah, good catch :)

src/lib/libreport-web.sym
msrb: This libtool magic was always confusing for me. How would you recommend updating the current values?
```
libreport_web_la_LDFLAGS = \
    -version-info 2:1:0
```

mgrabovsky: We have some instructions in the [wiki](https://github.com/abrt/libreport/wiki/Release-Checklist). Following the algorithm, I think the new version should be `3:0:1`.

msrb: I hope that won't require rebuild of abrt (?) :)

msrb: You're right. This blog post explains it quite nicely: https://begriffs.com/posts/2021-07-04-shared-libraries.html

I like semver better :)

msrb:
@mgrabovsky code updated :wink:

msrb:
I tested the updated code end-to-end and everything seems to be working well. Let's get this in :wink:

mgrabovsky:
Looks good. Sorry, I forgot to seal my approval :seal:

PR #753, opened 2022-07-25
abrt_xmlrpc: Don't warn about discarded const

michalfabik:
See commit. Just a warning I keep running into


mgrabovsky (approved):
Makes sense.

PR #758, opened 2022-09-14
reporter_bugzilla.py: Prevent possible KeyError exception

msrb:


mgrabovsky (approved):
Makes sense.

PR #759, opened 2022-09-14
More bugzilla-reporter fixes

msrb:


michalfabik:
Makes sense, thanks.

src/plugins/python/reporter_bugzilla.py
```diff
@@ -117,6 +117,8 @@ def set_settings(bz: Dict, settings: Dict):
         environ = settings.get('DontMatchComponents')
     bz['b_DontMatchComponents'] = environ or ''
 
+    if bz.get('b_create_private'):
+        bz['b_create_private'] = False
```
mgrabovsky, line 121: “If `b_create_private` is already set, override it to `False`”? Is this intended?

msrb: Ouch, no, of course not. I am missing "not" there. Good catch!

msrb: https://github.com/abrt/libreport/pull/760

PR #762, opened 2022-09-14
reporter_bugzilla.py: Don't ask for API key if we already have it

msrb:
set_settings() populates "rhbz" dict with settings from both config files and environment variables. gnome-abrt passes the API key to reporter-bugzilla via "Bugzilla_APIKey" env. variable.


michalfabik:
Looks good to me.

PR #764, opened 2022-09-19
Prevent few possible exceptions

msrb:

### DIFF
--- a/src/plugins/python/reporter_bugzilla.py
+++ b/src/plugins/python/reporter_bugzilla.py
@@ -848,7 +848,10 @@ if __name__ == '__main__':
 
         dup_comment = False
         trimmed_comment = re.sub(r'[ \t\n]', r'', bzcomment)
-        for c in bug_info['bi_comments']:
+
+        # We use "or []" here, because the value of "bi_comments"
+        # can be None and NoneType is not iterable.
+        for c in bug_info.get('bi_comments') or []:
             trimmed_c = re.sub(r'[ \t\n]', r'', c)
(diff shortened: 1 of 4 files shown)


src/plugins/python/reporter_bugzilla.py
```diff
@@ -848,7 +848,7 @@ def log_out(bug_info, rhbz, dump_dir_name):
 
         dup_comment = False
         trimmed_comment = re.sub(r'[ \t\n]', r'', bzcomment)
-        for c in bug_info['bi_comments']:
+        for c in bug_info.get('bi_comments') or []:
```
mgrabovsky, line 851: Why not just `bug_info.get('bi_comments', [])`?

msrb: Because the key `bi_comments` can exist, but the value can be `None`. The default value would be used only if the key doesn't exist.

mgrabovsky: Ah, OK. Could you perhaps add a short comment noting that `bi_comments` can be `None` and hence the `or`?

msrb: Good idea ;) I will probably get to that tomorrow though. No rush here :)

msrb: Updated (finally) ;)

msrb:
I've force-pushed so I'd appreciate a new review. Thanks! :)

mgrabovsky (approved):
Looks good to me.

PR #781, opened 2023-03-03
reporter-bugzilla: Fix password prompt in client/server mode

msrb:
Fixes: rhbz#2172891


msrb:
@michalfabik @mgrabovsky PTAL ;)

mgrabovsky:
Why not use the [`ask_password` function](https://github.com/abrt/libreport/blob/master/src/client-python/reportclient/clientmodule.c#L28) from the reportclient module?

msrb:
> Why not use the [ask_password function](https://github.com/abrt/libreport/blob/master/src/client-python/reportclient/clientmodule.c#L28) from the reportclient module?

Well, that's a good point. We don't use those bindings in the current reporter-bugzilla implementation. We only depend on satyr for low level stuff. This would be the only dependency on the old libreport C code and I'd prefer to avoid that, if possible.

michalfabik:
Looks fine to me.

mvollmer:
Thanks! This works, but there are still messages from getpass delivered with the Progress signal:

```
# python3 /var/tmp/BUG.py
/org/freedesktop/reportd/Task
PROGRESS Generating backtrace
PROGRESS Backtrace is generated and saved, 14455 bytes
PROMPT API key is not provided by configuration. Please enter the API key for 'https://bugzilla.redhat.com': 
PROGRESS /usr/lib64/python3.11/getpass.py:91: GetPassWarning: Can not control echo on the terminal.
PROGRESS   passwd = fallback_getpass(prompt, stream)
PROGRESS Warning: Password input may be echoed.
PROGRESS Password: 
...
```

This is mostly harmless, I guess, but a bit ugly.

PR #783, opened 2023-03-04
use $XDG_CONFIG_HOME to access user's configuration files (fix #782)

### LINKED ISSUE #782: Incorrect user configuration path computation in reporter_bugzilla.py
Python reporter-bugzilla fails to access `~/.config/libreport/bugzilla.conf` because `os.path.join()` discards the result of `os.environ['HOME']` as `USER_HOME_CONFIG_PATH` starts with a `/` (and `os.path.join()` assumes it's an absolute path).

Reported as [bug #2175415](https://bugzilla.redhat.com/show_bug.cgi?id=2175415) on Fedora.

ydroneaud:
Honor XDG_CONFIG_HOME variable if set, to access user's local configuration file, and fixes reporter_bugzilla.py's user local path computation.

For C code, Glib's g_get_user_config_dir() deals with XDG_CONFIG_HOME transparently for us.

For Python, external dependencies such as pyxdg, xdg-base-dirs, or appdirs could have been used to achieve the same behavior as Glib's g_get_user_config_dir(), but the choice to not depends on external packages was made.

Fixes #782.

### DIFF
--- a/src/client-python/reportclient/internal/const.py
+++ b/src/client-python/reportclient/internal/const.py
@@ -17,7 +17,6 @@
 
 GETTEXT_PROGNAME = 'reporter-bugzilla'
 CONF_DIR = '/etc/libreport'
-USER_HOME_CONFIG_PATH = '/.config/libreport'
 CONF_DIR_FLAG_NONE = 0
 CONF_DIR_FLAG_OPTIONAL = 1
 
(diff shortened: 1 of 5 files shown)


msrb (approved):

msrb:
Thank you! ;)

PR #795, opened 2023-06-30
Various fixes

msrb:
Multiple small changes that accumulated on my laptop overtime... :)

- add "XDG_ACTIVATION_TOKEN" to the list of ignored words (see the commit message for more details)
- tell Bugzilla that adding a attachment is a minor update so it doesn't send emails
- Fix NameError in reporter-bugzilla
- Retry Bugzilla queries with delays, if needed

### DIFF
--- a/src/client-python/reportclient/internal/bz_connection.py
+++ b/src/client-python/reportclient/internal/bz_connection.py
@@ -23,6 +23,7 @@ import logging
 import os
 import re
 import sys
+import urllib.parse
 from typing import Any, Dict, List, Optional
 
 import requests
@@ -363,7 +364,7 @@ class BZConnection:
         session.mount('http://', adapter)
         session.mount('https://', adapter)
         try:
-            response = session.get(os.path.join(self.url, 'rest.cgi/bug'),
+            response = session.get(urllib.parse.urljoin(self.url, 'rest.cgi/bug'),
                                     headers=self.headers,
                                     params=params,
                                     verify=self.verify)
--- a/src/client-python/reportclient/internal/bz_connection.py
+++ b/src/client-python/reportclient/internal/bz_connection.py
@@ -149,7 +149,8 @@ class BZConnection:
         data = json.dumps({'data': file_content,
                            'file_name': file_name,
                            'content_type': content_type,
-                           'summary': f"File: {file_name}"})
+                           'summary': f"File: {file_name}",
+                           'minor_update': True})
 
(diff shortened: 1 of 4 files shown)


src/client-python/reportclient/internal/bz_connection.py
```diff
@@ -26,6 +26,8 @@
 from typing import Any, Dict, List, Optional
 
 import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
```
mgrabovsky, line 30: urllib3 is not listed as a dependency in the spec file. Should it be added?

src/client-python/reportclient/internal/bz_connection.py
```diff
@@ -352,10 +354,22 @@ def bug_search(self, query: Dict) -> List:
...
-                                verify=self.verify)
+
+        # Configure requests to retry with delays (See: rhbz#2208742)
+        session = requests.Session()
+        retry = Retry(connect=5, backoff_factor=0.5)
+        adapter = HTTPAdapter(max_retries=retry)
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+        try:
+            response = session.get(os.path.join(self.url, 'rest.cgi/bug'),
```
mgrabovsky, line 366: This should be using `urllib.parse.urljoin()` rather than `os.path.join()`.

src/client-python/reportclient/internal/bz_connection.py
msrb: Good catch. `urllib3` is a dependency of [requests](https://github.com/psf/requests/blob/main/setup.py#L64) and that's probably why I haven't seen it explicitly listed in requirements very often.
But I actually wanted to use the built-in version that comes with Python itself... let me check what can be done here.

msrb: Yeah, it would be better to explicitly list `urllib3` as a a dependency: https://bugzilla.redhat.com/show_bug.cgi?id=1785666

msrb: I added both `requests` and `urllib3` to the spec file.

src/client-python/reportclient/internal/bz_connection.py
msrb: Fixed ;)

msrb:
@mgrabovsky Updated ;)

mgrabovsky (approved):
Looks good.

PR #802, opened 2024-01-21
Fix pointer type compatibility issues in tests

fweimer-rh:
Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


msrb:
Thank you! ;)

